PR #20, opened 2020-12-17
Support Resource Functions Returning Union with Error Type

ThisaruGuruge:
## Purpose
With this PR, the Ballerina GraphQL resources can return union types (with `error` type). 

This means the following resource function is allowed:

```ballerina
resource function get profile() returns string|error {

}
```

But the following is not allowed, since the union is only allowed with the error type.

```ballerina
resource function get id() returns int|string{

}
```

Fixes: [#762](https://github.com/ballerina-platform/ballerina-standard-library/issues/762)

### DIFF
--- a/graphql-ballerina/records.bal
+++ b/graphql-ballerina/records.bal
@@ -63,13 +63,22 @@ public type __Schema record {|
 #
 # + kind - The `graphql:__TypeKind` type of the type
 # + name - The name of the type
-# + fields - The fields of the given type
+# + fields - The fields of the given type, if the type qaulifies to have fields
+# + enumValues - The possible set of values, if the type is an enum
 public type __Type record {
     __TypeKind kind;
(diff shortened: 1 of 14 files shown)


graphql-ballerina/records.bal
```diff
@@ -101,17 +110,10 @@ public type __InputValue record {|
...
-    //INTERFACE,
-    //UNION,
-    //INPUT_OBJECT,
-    //LIST,
-    //NON_NULL
-}
+
+
+# Represents the type kind of a GraphQL type.
+public type __TypeKind "SCALAR"|"OBJECT"|"ENUM"|"NON_NULL";
```
ThisaruGuruge, line 116: This is changed since we cannot get the name of the `enum` type due to a limitation.

shafreenAnfar (approved):

PR #52, opened 2021-01-27
Publish Code Coverage Exec File as a Maven Artifact

ThisaruGuruge:
## Purpose
In order to check the code coverage of each of the module, we have to publish the `exec` file generated by the testerina. With this PR, the generated exec file will be named as `<module_name>-ballerina-<module_version>-<git_commit_hash>.exec`, and published to the GitHub packages as a separate maven artifact.

### DIFF
--- a/.github/workflows/build-master.yml
+++ b/.github/workflows/build-master.yml
@@ -33,12 +33,6 @@ jobs:
                     name: Ballerina Internal Log
                     path: graphql-ballerina/ballerina-internal.log
                     if-no-files-found: ignore
-            -   name: Archive Code Coverage JSON
-                uses: actions/upload-artifact@v2
-                with:
-                    name: Code Coverage JSON
-                    path: graphql-ballerina/target/report/test_results.json
-                    if-no-files-found: ignore
             -   name: Dispatch Dependent Module Builds
                 if: github.event.action != 'stdlib-publish-snapshot'
                 run: |
--- a/.github/workflows/build-master.yml
+++ b/.github/workflows/build-master.yml
@@ -33,6 +33,12 @@ jobs:
                     name: Ballerina Internal Log
                     path: graphql-ballerina/ballerina-internal.log
                     if-no-files-found: ignore
+            -   name: Archive Code Coverage JSON
+                uses: actions/upload-artifact@v2
+                with:
+                    name: Code Coverage JSON
+                    path: graphql-ballerina/target/report/test_results.json
+                    if-no-files-found: ignore
             -   name: Dispatch Dependent Module Builds
                 if: github.event.action != 'stdlib-publish-snapshot'
                 run: |
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -23,6 +23,12 @@ jobs:
                     name: Ballerina Internal Log
                     path: graphql-ballerina/ballerina-internal.log
                     if-no-files-found: ignore
+            -   name: Archive Code Coverage JSON
+                uses: actions/upload-artifact@v2
+                with:
+                    name: Code Coverage JSON
+                    path: graphql-ballerina/target/report/test_results.json
+                    if-no-files-found: ignore
 
     windows-build:
         runs-on: windows-latest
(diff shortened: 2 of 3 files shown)


graphql-ballerina/build.gradle
```diff
@@ -159,23 +158,29 @@ test {
...
+    }
+
     doLast {
         // Build and populate caches
         exec {
             workingDir project.projectDir
             environment "JAVA_OPTS", "-DBALLERINA_DEV_COMPILE_BALLERINA_ORG=true"
             if (Os.isFamily(Os.FAMILY_WINDOWS)) {
-                commandLine 'cmd', '/c', "$distributionBinPath/bal.bat build --skip-tests -c"
+                commandLine 'cmd', '/c', "$balJavaDebugParam $distributionBinPath/bal.bat build $testParams ${debugParams} && exit %%ERRORLEVEL%%"
```
azinneera, line 181: We can keep the `-c` flag. I think it's there to avoid creating the executable jar.

.github/workflows/pull-request.yml
```diff
@@ -23,12 +23,6 @@ jobs:
                     name: Ballerina Internal Log
                     path: graphql-ballerina/ballerina-internal.log
                     if-no-files-found: ignore
-            -   name: Archive Code Coverage JSON
```
azinneera, line 26: Do we need to remove this? IMO, keeping this will help developers get the exact lines for the coverage since we would not produce that info with jacoco report.

.github/workflows/build-master.yml
```diff
@@ -33,12 +33,6 @@ jobs:
                     name: Ballerina Internal Log
                     path: graphql-ballerina/ballerina-internal.log
                     if-no-files-found: ignore
-            -   name: Archive Code Coverage JSON
```
azinneera, line 36: Do we need to remove this? IMO, keeping this will help developers get the exact lines for the coverage since we would not produce that info with jacoco report.

graphql-ballerina/build.gradle
azinneera: If we are merging this after the lang repo PR merged, then we can keep the `--include=*` flag added to the command

graphql-ballerina/build.gradle
```diff
@@ -159,23 +158,29 @@ test {
...
+        }
+    }
+
     doLast {
         // Build and populate caches
         exec {
             workingDir project.projectDir
             environment "JAVA_OPTS", "-DBALLERINA_DEV_COMPILE_BALLERINA_ORG=true"
             if (Os.isFamily(Os.FAMILY_WINDOWS)) {
-                commandLine 'cmd', '/c', "$distributionBinPath/bal.bat build --skip-tests -c"
```
azinneera, line 176: There is another task to run tests AFAIK. Do we need to update this part in that case?

azinneera:
If we are merging this PR after the PR sent to lang-repo, then we can have the `--includes=*` added in the test command at https://github.com/ballerina-platform/module-ballerina-graphql/blob/3aefd771a665a691be83bf184e9c052e1218c02b/graphql-ballerina/build.gradle#L147 and L149

.github/workflows/build-master.yml
ThisaruGuruge: Fixed with https://github.com/ballerina-platform/module-ballerina-graphql/pull/52/commits/3aefd771a665a691be83bf184e9c052e1218c02b

graphql-ballerina/build.gradle
ThisaruGuruge: If we skip tests here, the build will run the test task first, and then run the build step after that. When we skip tests in build step, the code-coverage files will be lost. To handle that, we only run the test task if the build task is not present. (In line no. 166).

This way, we can preserve the code coverage files, and also keep the build as it is.

ThisaruGuruge:
> If we are merging this PR after the PR sent to lang-repo, then we can have the `--includes=*` added in the test command at
> 
> https://github.com/ballerina-platform/module-ballerina-graphql/blob/3aefd771a665a691be83bf184e9c052e1218c02b/graphql-ballerina/build.gradle#L147
> 
> and L149

@azinneera Will do

manuranga (approved):

PR #88, opened 2021-02-18
Re-order Build Tasks to Publish to the Ballerina Central

ThisaruGuruge:
## Purpose
$Subject

### DIFF
--- a/graphql-ballerina/build.gradle
+++ b/graphql-ballerina/build.gradle
@@ -137,8 +137,10 @@ task initializeVariables {
     }
 
     gradle.taskGraph.whenReady { graph ->
-        if (graph.hasTask(":${packageName}-ballerina:build") || graph.hasTask(":${packageName}-ballerina:publish")) {
+        if (graph.hasTask(":${packageName}-ballerina:build")) {
             ballerinaTest.enabled = false
+        } else {
+            ballerinaTest.enabled = true
         }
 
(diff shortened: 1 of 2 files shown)


aashikam (approved):

graphql-ballerina/build.gradle
```diff
@@ -147,8 +149,9 @@ task initializeVariables {
             testParams = "--skip-tests"
         }
 
-        if (graph.hasTask(":${packageName}-ballerina:publish")) {
+        if (graph.hasTask(":${packageName}-ballerina:publish") || graph.hasTask(":${packageName}-ballerina:publishToMavenLocal")) {
```
BuddhiWathsala, line 152: According to this logic, let say, someone has configured the ballerina central token. Then during the local publishToMavenLocal, the module will publish to central. What is the reason for doing that?.

ThisaruGuruge: We added this to enable publishing locally, when they have configured the access token. (Usually we don't). But I see your point.

Will come up with a solution.

shafreenAnfar: As you all discussed, I think we don't need to be able to publish to central from local machine. It is always good to do it via github actions.

ThisaruGuruge: Fixed with #89

PR #91, opened 2021-02-18
Avoid publishing to central fro SNAPSHOT versions

ThisaruGuruge:
## Purpose
$Subject

### DIFF
--- a/graphql-ballerina/build.gradle
+++ b/graphql-ballerina/build.gradle
@@ -220,6 +220,9 @@ task ballerinaBuild {
 
 task ballerinaPublish {
     doLast {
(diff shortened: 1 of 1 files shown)


graphql-ballerina/build.gradle
```diff
@@ -220,6 +220,9 @@ task ballerinaBuild {
 
 task ballerinaPublish {
     doLast {
+        if (!project.version.endsWith('-SNAPSHOT')) {
+            return
+        }
```
aashikam: Shouldn't this be 
```suggestion
        if (project.version.endsWith('-SNAPSHOT')) {
            return
        }
```

aashikam (approved):

PR #205, opened 2021-05-10
Fix Incorrect Location in Fragment Error Response

DimuthuMadushan:
## Purpose

Fixes: [ #1331 ](https://github.com/ballerina-platform/ballerina-standard-library/issues/1331)

## Examples
N/A
## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests


graphql-ballerina/modules/parser/parser.bal
```diff
@@ -127,15 +128,17 @@ public class Parser {
                     Selection selection = {
                         name: name,
                         isFragment: true,
-                        location: location
+                        location: location,
+                        ellipsisLocation: ellipsisLocation
```
ThisaruGuruge, line 132: We should not special case this. The `location` value of the selection should be the `ellipsisLocation` value when the selection is a fragment. Adding another field is not ideal. 

Is there a special reason to add this as a separate field?

graphql-ballerina/validator_visitor.bal
```diff
@@ -96,7 +96,7 @@ class ValidatorVisitor {
                     self.visitFragment(fragmentNode, fragmentParent);
                 } else {
                     string message = getFragmetCannotSpreadError(fragmentNode, selection.name, parentType);
-                    self.errors.push(getErrorDetailRecord(message, selection.location));
+                    self.errors.push(getErrorDetailRecord(message, selection?.spreadLocation ?: selection.location));
```
ThisaruGuruge, line 99:
```suggestion
                    self.errors.push(getErrorDetailRecord(message, <Location>selection?.spreadLocation));
```

graphql-ballerina/validator_visitor.bal
```diff
@@ -276,7 +276,7 @@ class ValidatorVisitor {
             __Type ofType = getOfType(schemaType);
             if (fragmentOnType != ofType) {
                 string message = getFragmetCannotSpreadError(fragmentNode, fragment.name, ofType);
-                ErrorDetail errorDetail = getErrorDetailRecord(message, fragment.location);
+                ErrorDetail errorDetail = getErrorDetailRecord(message, fragment?.spreadLocation ?: fragment.location);
```
ThisaruGuruge, line 279: Here, (also in the above scenario), we know that the selection has a `spreadLocation` value right? In that case, we can directly cast it and use it, instead of the condition. WDYT?
```suggestion
                ErrorDetail errorDetail = getErrorDetailRecord(message, <Location>fragment?.spreadLocation);
```

PR #216, opened 2021-05-13
Fix Query Depth Limit Ignoring in Fragments

DimuthuMadushan:
## Purpose

Fixes: 
[#1312](https://github.com/ballerina-platform/ballerina-standard-library/issues/1312)
## Examples
N/A
## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests


graphql-ballerina/query_depth_validator.bal
```diff
@@ -0,0 +1,103 @@
...
+                self.errors.push(getErrorDetailRecord(message, operationNode.getLocation()));
+                return;
+            }
+        }
+    }
+
+    public isolated function visitOperation(parser:OperationNode operationNode) {
+        foreach parser:Selection selection in operationNode.getSelections() {
+            self.visitSelection(selection);
+            if(self.queryDepth > self.maxQueryDepth) {
```
BuddhiWathsala, line 55: Please check with the Ballerina formatter. There should be a space after if. Further, we can remove the parenthesis of the if block.

graphql-ballerina/engine.bal
```diff
@@ -77,7 +77,14 @@ class Engine {
     }
 
     isolated function validateDocument(parser:DocumentNode document) returns OutputObject? {
-        ValidatorVisitor validator = new(<__Schema>self.schema, document, self.maxQueryDepth);
+        if (self.maxQueryDepth != 0) {
```
ThisaruGuruge, line 80:
```suggestion
        if (self.maxQueryDepth > 0) {
```

graphql-ballerina/query_depth_validator.bal
```diff
@@ -0,0 +1,103 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+import graphql.parser;
+
+class QueryDepthValidator{
+    private parser:DocumentNode documentNode;
+    private int queryDepth = 1;
+    private int maxQueryDepth = 0;
+    private int queryDepthLimit;
```
ThisaruGuruge, line 23: What's the difference between the `maxQueryDepth` and the `queryDepthLimit`?

graphql-ballerina/query_depth_validator.bal
```diff
@@ -0,0 +1,103 @@
...
+        if (self.errors.length() > 0) {
+            return self.errors;
+        }
+    }
+
+    public isolated function visitDocument(parser:DocumentNode documentNode) {
+        parser:OperationNode[] operations = documentNode.getOperations();
+        foreach parser:OperationNode operationNode in operations {
+            self.visitOperation(operationNode);
+            if (self.maxQueryDepth > self.queryDepthLimit) {
```
ThisaruGuruge, line 43: How could this condition become true?
If we want to stop the other operations from validating when we find an operation that exceeds the `maxQueryDepth`, we can use a boolean class variable, to set the **"mode"** of the visitor.
For example, say a variable `isMaxDepthExceeded`, which should be set to `false` at the init function and then set to `true` as soon as we find a query that exceeds the maximum depth. 

(But I think we should process the next operations as well and create errors for each operation.

graphql-ballerina/query_depth_validator.bal
```diff
@@ -0,0 +1,103 @@
...
+                self.errors.push(getErrorDetailRecord(message, operationNode.getLocation()));
+                return;
+            }
+        }
+    }
+
+    public isolated function visitOperation(parser:OperationNode operationNode) {
+        foreach parser:Selection selection in operationNode.getSelections() {
+            self.visitSelection(selection);
+            if (self.queryDepth > self.maxQueryDepth) {
```
ThisaruGuruge, line 55: Same comment as the above.

graphql-ballerina/query_depth_validator.bal
```diff
@@ -0,0 +1,103 @@
...
+                self.maxQueryDepth = self.queryDepth;
+            }
+            self.queryDepth = 0;
```
ThisaruGuruge, line 58: We start with `1` and here setting it back to `0`, doesn't seems right. The ideal way is to start from `0`, and increment when we walk into the operation node and decrement it when we walk out.

graphql-ballerina/query_depth_validator.bal
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+
+    public isolated function visitSelection(parser:Selection selection) {
+        if (selection.isFragment) {
+            parser:FragmentNode fragmentNode = self.documentNode.getFragments().get(selection.name);
+            if (fragmentNode.getFields().length() != 0 || fragmentNode.getFragments().length() != 0) {
```
ThisaruGuruge, line 65: Do we need this check here? If this is a fragment, it must have a set of selections, right? (Parser will make sure to fail if a fragment is not having a set of selections).

Ideally, when we find a fragment, we should visit that, (fragments do not increase the depth), and then process the selections of each fragment there.

graphql-ballerina/query_depth_validator.bal
```diff
@@ -0,0 +1,103 @@
...
+                self.visitFragment(fragmentNode);
+            }
+        } else {
+            parser:FieldNode fieldNode = <parser:FieldNode>selection?.node;
+            if (fieldNode.getFields().length() != 0 || fieldNode.getFragments().length() != 0) {
```
ThisaruGuruge, line 70: This check is also not necessary. We should visit the field right away. Inside the `visitField` function, we should check the conditions of the `fieldNode`.

graphql-ballerina/query_depth_validator.bal
```diff
@@ -0,0 +1,103 @@
...
+                self.queryDepth += 1;
```
ThisaruGuruge, line 71: Incrementing the query depth should be done in the `visitField` function. (Fields are the ones that increase the depth.)

graphql-ballerina/query_depth_validator.bal
```diff
@@ -0,0 +1,103 @@
...
+
+    public isolated function visitField(parser:FieldNode fieldNode) {
+        parser:Selection[] selections = fieldNode.getSelections();
+        int i = self.queryDepth;
+        foreach parser:Selection subSelection in selections {
+            self.visitSelection(subSelection);
+            if (self.queryDepth > self.maxQueryDepth) {
+                self.maxQueryDepth = self.queryDepth;
+            }
+            self.queryDepth = i;
```
ThisaruGuruge, line 85: If you increase the query depth when we walk into a `fieldNode`, then we don't need this. We just have to decrement it by `1` when we walk out from the field.

graphql-ballerina/query_depth_validator.bal
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+
+    public isolated function visitFragment(parser:FragmentNode fragmentNode) {
+        int i = self.queryDepth;
+        foreach parser:Selection selection in fragmentNode.getSelections() {
+            self.visitSelection(selection);
+            if (self.queryDepth > self.maxQueryDepth) {
```
ThisaruGuruge, line 93: We don't need this condition.

graphql-ballerina/query_depth_validator.bal
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+
+    public isolated function visitDocument(parser:DocumentNode documentNode) {
+        parser:OperationNode[] operations = documentNode.getOperations();
+        foreach parser:OperationNode operationNode in operations {
+            self.visitOperation(operationNode);
+            if (self.maxQueryDepth > self.queryDepthLimit) {
+                string message = string
+                `Query has depth of ${self.maxQueryDepth}, which exceeds max depth of ${self.queryDepthLimit}`;
```
ThisaruGuruge, line 45: We should improve this to add the query name (if exists) here to avoid ambiguities. Let's check whether the operation is not an anonymous operation, and if that's the case, let's add the operation name in the error message.

(Let's add a test for this too)

graphql-ballerina/validator_visitor.bal
```diff
@@ -21,14 +21,13 @@ class ValidatorVisitor {
 
     private __Schema schema;
     private parser:DocumentNode documentNode;
-    private int maxQueryDepth;
+    private int queryDepth = 0;
```
ThisaruGuruge, line 24: Why do we need this variable?

graphql-ballerina/query_depth_validator.bal
DimuthuMadushan: maxQueryDepth - maximum depth of the query that received.
queryDepthLimit - User given Limit of Depth that a query can have.

graphql-ballerina/validator_visitor.bal
DimuthuMadushan: To keep query depth temporary.

graphql-ballerina/query_depth_validator.bal
```diff
@@ -0,0 +1,101 @@
...
+
+    public isolated function visitDocument(parser:DocumentNode documentNode) {
+        parser:OperationNode[] operations = documentNode.getOperations();
+        foreach parser:OperationNode operationNode in operations {
+            self.visitOperation(operationNode);
+            if (self.maxQueryDepth > self.queryDepthLimit) {
+                if (operationNode.getName() != parser:ANONYMOUS_OPERATION) {
+                    string message = string
+                    `Query has depth of ${self.maxQueryDepth}, which exceeds max depth of ${self.queryDepthLimit} on ${
+                        operationNode.getName()}`;
```
ThisaruGuruge, line 47: This will show the error message in two lines. If you have a new line character in the middle of a string literal, the new line will also become a part of the string. So in this case, we may have to break the 120 rule.

Also, IIRC, this isn't the correct error message format, or am I missing out?

```suggestion
                    `Query "${operationNode.getName()}" has depth of ${self.maxQueryDepth}, which exceeds max depth of ${self.queryDepthLimit}`;
```

graphql-ballerina/query_depth_validator.bal
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    public isolated function visitField(parser:FieldNode fieldNode) {
+        self.queryDepth += 1;
+        if (fieldNode.getSelections().length() > 0) {
+            parser:Selection[] selections = fieldNode.getSelections();
+            foreach parser:Selection subSelection in selections {
+                self.visitSelection(subSelection);
+            }
+            self.queryDepth -= 1;
```
ThisaruGuruge, line 83: We can move this line out of the `if-else` blocks since this is common for both, and is duplicated.

graphql-ballerina/query_depth_validator.bal
```diff
@@ -0,0 +1,101 @@
...
+                    `Query has depth of ${self.maxQueryDepth}, which exceeds max depth of ${self.queryDepthLimit} on ${
+                        operationNode.getName()}`;
+                    self.errors.push(getErrorDetailRecord(message, operationNode.getLocation()));
+                } else {
+                    string message = string
+                    `Query has depth of ${self.maxQueryDepth}, which exceeds max depth of ${self.queryDepthLimit}`;
+                    self.errors.push(getErrorDetailRecord(message, operationNode.getLocation()));
+                }
+            }
+            self.queryDepth = 0;
```
ThisaruGuruge, line 55: Why do we have to set these back?
Once we are done with the document node, we are done, right?

graphql-ballerina/query_depth_validator.bal
```diff
@@ -0,0 +1,101 @@
...
+            self.maxQueryDepth = 0;
+        }
+    }
+
+    public isolated function visitOperation(parser:OperationNode operationNode) {
```
ThisaruGuruge, line 60: Shouldn't `visitOperation` increase the `queryDepth` by 1?

graphql-ballerina/tests/11_listener_configurations_test.bal
```diff
@@ -122,4 +135,69 @@ isolated function testQueryExceedingMaxDepth() returns error? {
...
+            }
+        ]
+    };
+    test:assertEquals(actualPayload, expectedPayload);
+}
+
+@test:Config {
+    groups: ["configs", "unit"]
+}
+isolated function testQueryWithNamedOperationExceedingMaxDepth() returns error? {
```
ThisaruGuruge, line 175: Let's add a test with a document with more than one query.

graphql-ballerina/validator_visitor.bal
ThisaruGuruge: I don't see any usage of this variable here.

graphql-ballerina/query_depth_validator.bal
```diff
@@ -0,0 +1,103 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import graphql.parser;
+
+class QueryDepthValidator{
+    private parser:DocumentNode documentNode;
+    private int queryDepth = 1;
+    private int maxQueryDepth = 0;
```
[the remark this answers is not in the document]
DimuthuMadushan: maxQueryDepth is not the one we get from the engine. maxQueryDepth uses to keep the maximum depth of a document.

graphql-ballerina/query_depth_validator.bal
DimuthuMadushan: This is inside the ```for``` loop. It sets back after visiting an operation.

graphql-ballerina/query_depth_validator.bal
DimuthuMadushan: Operation must have a set of selection otherwise Parser will make sure to fail it. ```queryDepth``` is increased when visiting to that selections.

graphql-ballerina/query_depth_validator.bal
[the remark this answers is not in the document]
ThisaruGuruge: These should be initialized in the `init` function.

graphql-ballerina/query_depth_validator.bal
ThisaruGuruge: If we increase the `queryDepth` by `1` when we visiting an Operation, and decrement it when we walk out from the Operation, we do not need this.

graphql-ballerina/query_depth_validator.bal
ThisaruGuruge: I get that. But since we are starting the `queryDepth` from `0`, shouldn't we increase it by `1` when we visit an operation? Wouldn't this cause incorrect counts?

graphql-ballerina/tests/11_listener_configurations_test.bal
```diff
@@ -122,4 +135,90 @@ isolated function testQueryExceedingMaxDepth() returns error? {
...
+ }
+ query {
+    Student {
+        name
+    }
+ }`;
+    string url = "http://localhost:9103/depthLimitService";
+    json actualPayload = check getJsonPayloadFromService(url, document);
+    json expectedPayload = {
+        errors: [
```
ThisaruGuruge, line 169: We need to check the query names are correct with the positions. So we should use named operations here.

This should return another error, `This anonymous operation must be the only defined operation.`

A document cannot have multiple anonymous queries. If this does not return that error, shall we create an issue?

graphql-ballerina/validator_visitor.bal
ThisaruGuruge: This variable is not used. Shall we remove it?

ThisaruGuruge (approved):

graphql-ballerina/tests/11_listener_configurations_test.bal
DimuthuMadushan: ```query_depth_validator``` execute before the ```validator_visitor```. Therefore that error won't hit.

PR #275, opened 2021-06-09
Fix Fields Missing when a Query Having Duplicate Fields

DimuthuMadushan:
## Purpose

Fixes: [ #1277 ](https://github.com/ballerina-platform/ballerina-standard-library/issues/1277) 

## Examples
N/A
## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests


graphql-ballerina/duplicate_field_remover.bal
```diff
@@ -0,0 +1,181 @@
...
+        foreach int index in duplicateIndexes.reverse() {
+            var remove = selections.remove(index);
+        }
+    }
+
+    private isolated function removeDuplicateFields(parser:FieldNode[] fields) {
+        string[] visitedFields = [];
+        int[] duplicateIndexes = [];
+        int i = 0;
+        while (i < fields.length()) {
```
ThisaruGuruge, line 110: Can use a foreach loop here instead.

graphql-ballerina/duplicate_field_remover.bal
```diff
@@ -0,0 +1,181 @@
...
+            i += 1;
+        }
+        foreach int index in duplicateIndexes.reverse() {
+            var remove = selections.remove(index);
+        }
+    }
+
+    private isolated function removeDuplicateFields(parser:FieldNode[] fields) {
+        string[] visitedFields = [];
+        int[] duplicateIndexes = [];
```
ThisaruGuruge, line 108: We can make this a `parser:FieldNode[]` instead of an int array. Don't need to keep the indexes, instead keep the original object reference.

graphql-ballerina/duplicate_field_remover.bal
```diff
@@ -0,0 +1,181 @@
...
+        }
+    }
+
+    private isolated function removeDuplicateSelections(parser:Selection[] selections) {
+        map<parser:Selection> visitedSelections = {};
+        map<parser:Selection> visitedFragmentOnTypes = {};
+        int[] duplicateIndexes = [];
+        int i = 0;
+        while (i < selections.length()) {
+            parser:Selection subSelection = selections[i];
```
ThisaruGuruge, line 79: Usually the prefix `sub` is used when there is a parent of the same type that exists within the scope. Otherwise, there's no need to use `sub` here.

Check the other places as well.

```suggestion
            parser:Selection selection = selections[i];
```

graphql-ballerina/duplicate_field_remover.bal
```diff
@@ -0,0 +1,181 @@
...
+            if(subSelection.isFragment) {
+                parser:FragmentNode subFragmentNode = self.documentNode.getFragments().get(subSelection.name);
+                if (visitedFragmentOnTypes.hasKey(subFragmentNode.getOnType())) {
+                    if (subFragmentNode.isInlineFragment()) {
+                        duplicateIndexes.push(i);
+                    } else {
+                        self.appendDuplicateSelections(selections[i], visitedFragmentOnTypes.get(subFragmentNode.getOnType()));
```
ThisaruGuruge, line 86:
```suggestion
                        self.appendDuplicateSelections(subSelection, visitedFragmentOnTypes.get(subFragmentNode.getOnType()));
```

graphql-ballerina/duplicate_field_remover.bal
```diff
@@ -0,0 +1,181 @@
...
+            i += 1;
+        }
+        foreach int index in duplicateIndexes.reverse() {
+            var remove = fields.remove(index);
+        }
+    }
+
+    private isolated function removeDuplicateFragments(string[] fragments) {
+        string[] visitedFragmentOnTypes = [];
+        int[] duplicateIndexes = [];
```
ThisaruGuruge, line 126: Same as above.

graphql-ballerina/duplicate_field_remover.bal
```diff
@@ -0,0 +1,181 @@
...
+                if (visitedSelections.hasKey(subSelection.name)) {
+                    self.appendDuplicateSelections(selections[i], visitedSelections.get(subSelection.name));
+                    duplicateIndexes.push(i);
+                } else {
+                    visitedSelections[subSelection.name] = selections[i];
+                }
+            }
+            i += 1;
+        }
+        foreach int index in duplicateIndexes.reverse() {
```
ThisaruGuruge, line 101: Why do we have to `reverse` this here?

graphql-ballerina/duplicate_field_remover.bal
```diff
@@ -0,0 +1,181 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import graphql.parser;
+
+class DuplicateFieldRemover {
+    private parser:DocumentNode documentNode;
```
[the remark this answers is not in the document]
DimuthuMadushan: Since we are not going to use ```visitArgument``` function here, we can not extend the ```parser:Visitor``` class.

changelog.md
```diff
@@ -6,6 +6,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Added
+
+### Changed
+
```
ThisaruGuruge, line 12: Let's add these when needed. 

```suggestion
```

graphql-ballerina/duplicate_field_remover.bal
```diff
@@ -0,0 +1,194 @@
...
+                    self.appendDuplicateSelections(selection, visitedSelections.get(selection.name));
+                    duplicateSelections.push(selection);
+                } else {
+                    visitedSelections[selection.name] = selection;
+                }
+            }
+        }
+        int i = 0;
+        while (i < selections.length()) {
+            foreach parser:Selection duplicate in  duplicateSelections {
```
ThisaruGuruge, line 107:
```suggestion
            foreach parser:Selection duplicate in duplicateSelections {
```

graphql-ballerina/duplicate_field_remover.bal
```diff
@@ -0,0 +1,194 @@
...
+        while (i < fields.length()) {
+            foreach parser:FieldNode duplicate in  duplicateFields {
+                if (fields[i] === duplicate) {
+                    parser:FieldNode remove = fields.remove(i);
+                    i -= 1;
+                }
+            }
+            i += 1;
+        }
+    }
```
ThisaruGuruge, line 137: Can't we use a single loop here?

Maybe using a map to keep the duplicates? I feel like this logic can be improved to log(n)

graphql-ballerina/duplicate_field_remover.bal
```diff
@@ -0,0 +1,194 @@
...
+        foreach var selections in duplicate.getSelections() {
+            original.addSelection(selections);
+        }
+        foreach var arguments in duplicate.getArguments() {
+            original.addArgument(arguments);
+        }
+    }
+
+    private isolated function appendDuplicateFragments(parser:FragmentNode duplicate, parser:FragmentNode original) {
+        foreach var fields in duplicate.getFields() {
```
ThisaruGuruge, line 184: In these places it is better to use the exact type instead of `var`, since we do know the type of the element. Check the other places as well.

ThisaruGuruge:
Shall we check the reduced test coverage too?
Maybe a false positive, but better to check it.

graphql-ballerina/duplicate_field_remover.bal
DimuthuMadushan: Can not have a unique key for each duplicates field node. So that it's hard to use a map here.

DimuthuMadushan: Any suggestions?

ThisaruGuruge: Can't we use `arrays:filter()` here?

DimuthuMadushan: We need two loop if we use ```arrays:filter( )```, don't we? one for filter function and one for put filtered values into ```fields``` ?

ThisaruGuruge: This is what I have in mind:

```ballerina
private isolated function removeDuplicateFields(parser:ParentNode node) {
    parser:fieldNode[] fields = node.getFields();
    string[] visitedFields = [];
    parser:FieldNode[] duplicateFields = [];
    foreach parser:FieldNode fieldNode in fields {
        if (visitedFields.indexOf(fieldNode.getName()) == ()) {
             visitedFields.push(fieldNode.getName());
        } else {
            duplicateFields.push(fieldNode);
        }
    }
    int i = 0;

    // Filter the fieldsArray here, and assign it back to the node.
    node.fields = filteredArray 
}
```

You may have to introduce a new function in the `parentNode` class to set the fields.

ThisaruGuruge (approved):

PR #299, opened 2021-06-17
Fix Response Status Code for Responses with Errors

DimuthuMadushan:
## Purpose

Fixes: [ #1218 ](https://github.com/ballerina-platform/ballerina-standard-library/issues/1218)
Fixes: [ #1508 ](https://github.com/ballerina-platform/ballerina-standard-library/issues/1508)

## Examples
N/A
## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests


graphql-ballerina/engine.bal
```diff
@@ -27,22 +27,23 @@ isolated class Engine {
         self.maxQueryDepth = getMaxQueryDepth(serviceConfig);
     }
 
-    isolated function getOutputObjectForQuery(string documentString, string operationName) returns OutputObject {
+    isolated function getOutputObjectForQuery(string documentString, string operationName) returns
+                                                                                        [OutputObject, string] {
         parser:DocumentNode|OutputObject result = self.parse(documentString);
         if (result is OutputObject) {
-            return result;
+            return [result, "SyntaxError"];
```
ThisaruGuruge, line 34: It is better to define constants for the error codes and use them here instead of using strings directly. Otherwise, if we have to change the error code, we have to change every occurrence of this string in the code.

graphql-ballerina/engine.bal
```diff
@@ -27,22 +27,23 @@ isolated class Engine {
...
         }
         parser:DocumentNode document = <parser:DocumentNode>result;
         OutputObject? validationResult = self.validateDocument(document);
         if (validationResult is OutputObject) {
-            return validationResult;
+            return [validationResult, "DocumentError"];
         } else {
             if (document.getOperations().length() == 1) {
-                return self.execute(document.getOperations()[0]);
+                return [self.execute(document.getOperations()[0]), ""];
```
ThisaruGuruge, line 42: Better to use `()` here instead of an empty string.

graphql-ballerina/listener_utils.bal
```diff
@@ -26,7 +26,10 @@ isolated function handleGetRequests(Engine engine, http:Request request) returns
         setResponseForBadRequest(response);
     } else {
         string operationName = resolveOperationName(request.getQueryParamValue(PARAM_OPERATION_NAME));
-        OutputObject outputObject = engine.getOutputObjectForQuery(query, operationName);
+        [OutputObject, string] [outputObject, errorType] = engine.getOutputObjectForQuery(query, operationName);
+        if (errorType == "SyntaxError" || errorType == "DocumentError") {
+            response.statusCode = 400;
```
ThisaruGuruge, line 31: Shall we add a test for this scenario as well?

graphql-ballerina/listener_utils.bal
```diff
@@ -65,7 +68,10 @@ isolated function processJsonPayload(Engine engine, json payload, http:Response
         operationName = resolveOperationName(operationNameInPayload);
     }
     if (documentString is string) {
-        OutputObject outputObject = engine.getOutputObjectForQuery(documentString, operationName);
+        [OutputObject, string] [outputObject, errorType] = engine.getOutputObjectForQuery(documentString, operationName);
+        if (errorType == "SyntaxError" || errorType == "DocumentError") {
+            response.statusCode = 400;
+        }
```
ThisaruGuruge, line 74: Since setting the bad request status code is duplicated, shall we move it to a separate function and call it?

graphql-ballerina/tests/06_errors_test.bal
```diff
@@ -52,7 +56,10 @@ isolated function testMissingArgumentValueQuery() returns error? {
 isolated function testEmptyQuery() returns error? {
     string graphqlUrl = "http://localhost:9095/graphql";
     string document = "{ }";
-    json result = check getJsonPayloadFromService(graphqlUrl, document);
+    http:Client httpClient = check new(graphqlUrl);
+    http:Response response = check httpClient->post("/", { query: document });
+    test:assertEquals(response.statusCode, 400);
```
ThisaruGuruge, line 61: Since these lines are duplicated in a few places, we can move this to a separate utility function as well.

graphql-ballerina/tests/utils.bal
```diff
@@ -22,9 +22,11 @@ isolated function getJsonPayloadFromService(string url, string document, string?
 returns json|error {
     http:Client httpClient = check new(url);
     if (operationName is string) {
-        return check httpClient->post("/", { query: document, operationName: operationName }, targetType = json);
+        http:Response response = check httpClient->post("/", { query: document, operationName: operationName });
+        return response.getJsonPayload();
     }
-    return check httpClient->post("/", { query: document }, targetType = json);
+    http:Response response = check httpClient->post("/", { query: document });
+    return response.getJsonPayload();
```
ThisaruGuruge, line 29: What's the reason for this change? We can even simply this as follows:

```suggestion
        http:Client httpClient = check new(url);
        return httpClient->post("/", { query: document, operationName: operationName });
```

graphql-ballerina/listener_utils.bal
```diff
@@ -65,7 +68,10 @@ isolated function processJsonPayload(Engine engine, json payload, http:Response
         operationName = resolveOperationName(operationNameInPayload);
     }
     if (documentString is string) {
-        OutputObject outputObject = engine.getOutputObjectForQuery(documentString, operationName);
+        [OutputObject, string] [outputObject, errorType] = engine.getOutputObjectForQuery(documentString, operationName);
+        if (errorType == "SyntaxError" || errorType == "DocumentError") {
+            response.statusCode = 400;
```
ThisaruGuruge, line 73: Instead of setting the status code directly, I guess we can directly use the `http:BAD_REQUEST` response here, but I'm not sure. Can you check that as well?

graphql-ballerina/tests/utils.bal
DimuthuMadushan: If we return ```httpClient->post("/", { query: document, operationName: operationName })``` here, tests returning responses with errors will hit the ```check``` at ```json actualPayload = check getJsonPayloadFromService(url, document)```.

graphql-ballerina/engine.bal
```diff
@@ -51,7 +52,7 @@ isolated class Engine {
                 message: message,
                 locations: []
             };
-            return getOutputObjectFromErrorDetail(errorDetail);
+            return [getOutputObjectFromErrorDetail(errorDetail), DOCUMENT_ERROR];
```
ThisaruGuruge, line 55: This should be a different scenario (If we are to treat these as different scenarios).

There are two ways:
1. Treat all the errors that occurred before the execution as the same, (`VALIDATION_ERROR` maybe?). Then we can add that to all the places. Because we only need this information to decide whether to send `400` or `200` as the status code, I guess this approach is okay. In that case, we can even return a boolean value as the second element in the tuple. (`isValidationError` maybe?)
2. Treat different errors that occurred at different validation phases as different errors. AFAIU, we don't need that information (at least for now) to decide the status code for the response. But if we are to treat them as different errors, this is a separate error type. (`INVALID_REQUEST` maybe?)

WDYT?

graphql-ballerina/tests/utils.bal
```diff
@@ -17,17 +17,23 @@
...
 
 isolated function getJsonPayloadFromService(string url, string document, string? operationName = ())
 returns json|error {
     http:Client httpClient = check new(url);
-    if (operationName is string) {
-        return check httpClient->post("/", { query: document, operationName: operationName }, targetType = json);
-    }
-    return check httpClient->post("/", { query: document }, targetType = json);
+    http:Response response = check httpClient->post("/", { query: document, operationName: operationName });
+    return response.getJsonPayload();
```
ThisaruGuruge, line 26:
```suggestion
    return httpClient->post("/", { query: document, operationName: operationName });
```

graphql-ballerina/listener_utils.bal
```diff
@@ -95,3 +95,26 @@ isolated function resolveOperationName(string? operationName) returns string {
...
+isolated function setResponse(Engine engine, http:Response response, parser:OperationNode|OutputObject validationResult) {
+    if (validationResult is OutputObject) {
+        OutputObject outputObject = <OutputObject> validationResult;
+        json|error payload  = outputObject.cloneWithType();
+        if (payload is error) {
+            response.statusCode = http:STATUS_INTERNAL_SERVER_ERROR;
+            response.setJsonPayload(<json> payload.message());
+        } else {
+            response.statusCode = http:STATUS_BAD_REQUEST;
+            response.setJsonPayload(<json> payload);
```
ThisaruGuruge, line 108: This cast is unnecessary since type narrowing will narrow the type to `json`.

```suggestion
            response.setJsonPayload(payload);
```

graphql-ballerina/listener_utils.bal
```diff
@@ -95,3 +95,26 @@ isolated function resolveOperationName(string? operationName) returns string {
...
+        }
+    } else {
+        OutputObject outputObject = engine.execute(<parser:OperationNode> validationResult);
+        json|error payload  = outputObject.cloneWithType();
+        if (payload is error) {
+            response.statusCode = http:STATUS_INTERNAL_SERVER_ERROR;
+            response.setJsonPayload(<json> payload.message());
+        } else {
+            response.setJsonPayload(<json> payload);
```
ThisaruGuruge, line 117: Same as above.

```suggestion
            response.setJsonPayload(payload);
```

graphql-ballerina/tests/utils.bal
```diff
@@ -17,17 +17,22 @@
...
+import ballerina/test;
 
 isolated function getJsonPayloadFromService(string url, string document, string? operationName = ())
 returns json|error {
     http:Client httpClient = check new(url);
-    if (operationName is string) {
-        return check httpClient->post("/", { query: document, operationName: operationName }, targetType = json);
-    }
-    return check httpClient->post("/", { query: document }, targetType = json);
+    return  check httpClient->post("/", { query: document, operationName: operationName });
```
ThisaruGuruge, line 25:
```suggestion
    return httpClient->post("/", { query: document, operationName: operationName });
```

ThisaruGuruge (approved):

PR #344, opened 2021-07-21
Add Support for GraphQL Variables

DimuthuMadushan:
## Purpose

Fixes: [ #1361 ](https://github.com/ballerina-platform/ballerina-standard-library/issues/1361)
Fixes: [ #1688 ](https://github.com/ballerina-platform/ballerina-standard-library/issues/1688)

## Examples
N/A
## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests


ballerina/modules/parser/operation_node.bal
```diff
@@ -69,4 +73,22 @@ public class OperationNode {
     public isolated function getSelections() returns Selection[] {
         return self.selections;
     }
+
+    public isolated function addVariableDefinition(string name, VariableDefinition varDef) {
```
ThisaruGuruge, line 77: Can't we get the `name` from the `VariableDefinition` record itself?

ballerina/modules/parser/parser.bal
```diff
@@ -104,10 +103,56 @@ public class Parser {
...
+        if (token.kind == T_OPEN_PARENTHESES) {
+            check self.addVariableDefinition(operation);
+        }
         check self.addSelections(operation);
         return operation;
     }
 
+    isolated function addVariableDefinition(OperationNode operationNode) returns Error? {
+        Token token = check self.readNextNonSeparatorToken(); // Read the open parantheses here
+        while (token.kind != T_CLOSE_PARENTHESES) {
```
ThisaruGuruge, line 116: Let's remove the parenthesis.
```suggestion
        while token.kind != T_CLOSE_PARENTHESES {
```

ballerina/modules/parser/parser.bal
```diff
@@ -104,10 +103,56 @@ public class Parser {
...
+            token = check self.readNextNonSeparatorToken();
+            if (token.kind != T_DOLLAR) {
```
ThisaruGuruge, line 118: Let's do not add parentheses around the simple conditional expressions. Check the other places as well.

```suggestion
            if token.kind != T_DOLLAR {
```

ballerina/modules/parser/parser.bal
```diff
@@ -104,10 +103,56 @@ public class Parser {
...
+                operationNode.addVariableDefinition(variableName, varDefinition);
+                token = check self.peekNextNonSeparatorToken();
+            } else {
+                VariableDefinition varDefinition = {
+                    name: variableName,
+                    kind: varType,
+                    location: location
+                };
+                operationNode.addVariableDefinition(variableName, varDefinition);
+            }
```
ThisaruGuruge, line 151: This function is too long. Better to split this up.

ballerina/modules/parser/records.bal
```diff
@@ -29,6 +29,19 @@ public type Location record {|
     int column;
 |};
 
+# Store a Varibale definition in a GraphQL operation.
```
ThisaruGuruge, line 32:
```suggestion
# Stores a variable defined in a GraphQL operation.
```

ballerina/query_depth_validator.bal
```diff
@@ -67,8 +67,10 @@ class QueryDepthValidator{
 
     public isolated function visitSelection(parser:Selection selection, anydata data = ()) {
         if (selection.isFragment) {
-            parser:FragmentNode fragmentNode = self.documentNode.getFragments().get(selection.name);
-            self.visitFragment(fragmentNode);
+            if (self.documentNode.getFragments().keys().indexOf(selection.name) != ()) {
```
ThisaruGuruge, line 70: Instead of this, can't we validate the fragments first? Let's move the `FragmentVisitor` initialization and validation to `Engine` and then validate fragments first.

ballerina/tests/22_variables.bal
```diff
@@ -0,0 +1,220 @@
+// Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
ThisaruGuruge, line 1:
```suggestion
// Copyright (c) 2021, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```

ballerina/modules/parser/types.bal
```diff
@@ -27,7 +27,7 @@ type CharIterator object {
 type Boolean TRUE|FALSE;
 
 type SeparatorType T_EOF|T_WHITE_SPACE|T_COMMA|T_NEW_LINE;
-type SpecialCharacterType T_OPEN_BRACE|T_CLOSE_BRACE|T_OPEN_PARENTHESES|T_CLOSE_PARENTHESES|T_COLON|T_COMMENT;
+type SpecialCharacterType T_OPEN_BRACE|T_CLOSE_BRACE|T_OPEN_PARENTHESES|T_CLOSE_PARENTHESES|T_COLON|T_COMMENT|T_DOLLAR|T_EQUAL;
```
ThisaruGuruge, line 30: Let's add parser and lexer tests for these cases.

ballerina/tests/22_variables.bal
```diff
@@ -0,0 +1,220 @@
...
+    json actualPayload = check getJsonPayloadFromBadRequest(url, document, variables);
+    json expectedPayload = check getJsonContentFromFile("fragments_with_undefined_variables.json");
+    assertJsonValuesWithOrder(actualPayload, expectedPayload);
+}
+
+@test:Config {
+    groups: ["variables", "fragments", "input"]
+}
+isolated function testDuplicateVariablesWithMultipleOperations() returns error? {
+    string document = check getGraphQLDocumentFromFile("multiple_operations_with_duplicate_variables.txt");
```
ThisaruGuruge, line 95: This document file name and test name does not match.

ballerina/tests/22_variables.bal
```diff
@@ -0,0 +1,220 @@
...
+    json expectedPayload = check getJsonContentFromFile("variables_with_default_values.json");
+    assertJsonValuesWithOrder(actualPayload, expectedPayload);
+}
+
+@test:Config {
+    groups: ["variables", "inputs", "input_coerce"]
+}
+isolated function testVariablesWithCoerceIntInputToFloat() returns error? {
+    string document = "($weight:float){ weightInPounds(weightInKg:$weight) }";
+    json variables = { "weight": 1};
```
ThisaruGuruge, line 143: When defining a JSON map in Ballerina, you don't necessarily have to use double quotes for the keys.

```suggestion
    json variables = { weight: 1};
```

ballerina/tests/22_variables.bal
```diff
@@ -0,0 +1,220 @@
...
+    };
+    assertJsonValuesWithOrder(payloadWithFloatValues, expectedPayload);
+}
+
+@test:Config {
+    groups: ["variables", "fragments", "input"]
+}
+isolated function testVariablesWithMissingRequiredArgument() returns error? {
+    string document = string`query Greeting($userName:string){ greet (name: $userName ) }`;
+    string url = "http://localhost:9091/records";
```
ThisaruGuruge, line 160: This service does not have a field named `greet`?

ballerina/tests/22_variables.bal
```diff
@@ -0,0 +1,220 @@
...
+    json expectedPayload = check getJsonContentFromFile("variables_with_missing_required_argument.json");
+    assertJsonValuesWithOrder(actualPayload, expectedPayload);
+}
+
+@test:Config {
+    groups: ["variables", "inputs", "enums"]
+}
+isolated function testEnumTypeVariables() returns error? {
+    string document = "($day:Weekday){ isHoliday(weekday: $day) }";
+    json variables = { "day": MONDAY };
```
ThisaruGuruge, line 171:
```suggestion
    json variables = { day: MONDAY };
```

ballerina/tests/resources/expected_results/duplicate_input_variables.json
```diff
@@ -0,0 +1,13 @@
...
+            "message": "There can be only one variable named \"$userName\"",
+            "locations": [
+                {
+                    "line": 1,
+                    "column": 20
+                }
+            ]
+        }
+    ]
+}
```
ThisaruGuruge, line 13:
```suggestion
}

```

ballerina/tests/resources/expected_results/fragments_with_input_variables.json
```diff
@@ -0,0 +1,9 @@
+{
+    "data": {
+        "profile": {
+            "address": {
+                "city": "Albuquerque"
+            }
+        }
+    }
+}
```
ThisaruGuruge, line 9: Check the other places as well.
```suggestion
}

```

ballerina/tests/resources/expected_results/multiple_operations_with_duplicate_variables.json
```diff
@@ -0,0 +1,13 @@
+{
+    "errors": [
+        {
+            "message": "Variable \"$profileId\" got invalid value 1;Expected type String. String cannot represent value: 1",
```
ThisaruGuruge, line 4: This does not express the presence of a duplicate variable. It's just a type mismatch. We have to check the error when a duplicate variable is defined with the same type, I think.

ballerina/variable_validator.bal
```diff
@@ -0,0 +1,149 @@
...
+        } else {
+            string message = string`Variable "$${variableName}" is not defined.`;
+            self.errors.push(getErrorDetailRecord(message, location)); 
+        }
+    }
+
+    public isolated function setArgumentValue(Scalar value, parser:ArgumentNode argument, parser:Location location) {
+        if argument.getKind() == parser:T_IDENTIFIER {
+            argument.setValue(value);
+        } else if (typeof value).toString().equalsIgnoreCaseAscii("typedesc " + getTypeName(argument)) {
```
ThisaruGuruge, line 135: When comparing type names, let's not use `equalsIgnoreCase` function, since these names are case sensitive.

Shall we move retrieving the type name from the value to a separate function?

@MaryamZi aren't there any API to get the type name from a typedesc? If that's the case, shouldn't we have one to get type name from a typedesc value?

ThisaruGuruge:
Let's sync with the master branch.

ballerina/variable_validator.bal
```diff
@@ -0,0 +1,147 @@
...
+        }
+    }
+
+    public isolated function visitDocument(parser:DocumentNode documentNode, anydata data = ()) {
+        parser:OperationNode[] operations = documentNode.getOperations();
+        foreach parser:OperationNode operationNode in operations {
+            foreach ErrorDetail errorDetail in operationNode.getErrors() {
+                self.errors.push(errorDetail);
+            }
+            self.visitOperation(operationNode, operationNode.getVaribleDefinitions());
```
ThisaruGuruge, line 51: Why do we have to pass the variable definitions separately? We are passing the `OpearationNode`, the variable definitions are already passing, right?

ballerina/variable_validator.bal
```diff
@@ -0,0 +1,147 @@
...
+        }
+    }
+
+    public isolated function visitOperation(parser:OperationNode operationNode, anydata data = ()) {
+        map<parser:VariableDefinition> variableDefinitions = <map<parser:VariableDefinition>> data;
```
ThisaruGuruge, line 56: If we directly access the variable definitions from the operation node, this casting can be avoided.

ballerina/modules/parser/operation_node.bal
```diff
@@ -69,4 +73,22 @@ public class OperationNode {
...
+            Location location = <Location> varDef?.location;
+            self.errors.push({message: message, locations:[location]});
+        } else { 
+            self.variables[varDef.name] = varDef;
+        }
+    }
+
+    public isolated function getVaribleDefinitions() returns map<VariableDefinition> {
+        return self.variables;
+    }
```
ThisaruGuruge, line 89: Since we need one variable at a time, a better approach would be creating a `hasVariable` function and a `getVariable` function. This way, we can check whether a variable is defined and if it is defined get only that variable when needed. Otherwise, we have to get the map all time and process it every time.

ballerina/modules/parser/tests/parser_tests.bal
```diff
@@ -486,3 +486,31 @@ isolated function testFieldAliasWithArguments() returns error? {
...
+}
+isolated function testVariables() returns error? {
+    string document = "query getName($profileId:Int = 3) { profile(id:$profileId) { name } }";
+    Parser parser = new(document);
+    DocumentNode documentNode = check parser.parse();
+    OperationNode operationNode = <OperationNode>documentNode.getOperations()[0];
+    test:assertEquals(operationNode.getVaribleDefinitions().length(), 1);
+    VariableDefinition variableDefinition = <VariableDefinition> operationNode.getVaribleDefinitions()["profileId"];
+    string variableName = variableDefinition.name;
+    test:assertEquals(variableName, "profileId");
```
ThisaruGuruge, line 501: Let's refrain from craeting unnecessary variables.  
```suggestion
    test:assertEquals(variableDefinition.name, "profileId");
```

ballerina/query_depth_validator.bal
```diff
@@ -68,7 +68,7 @@ class QueryDepthValidator{
     public isolated function visitSelection(parser:Selection selection, anydata data = ()) {
         if (selection.isFragment) {
             parser:FragmentNode fragmentNode = self.documentNode.getFragments().get(selection.name);
-            self.visitFragment(fragmentNode);
+            self.visitFragment(fragmentNode, data);
```
ThisaruGuruge, line 71: Do we need this?

ballerina/tests/22_variables.bal
```diff
@@ -0,0 +1,220 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+import ballerina/test;
+import ballerina/lang.value;
+@test:Config {
+    groups: ["variables", "input"]
+}
+isolated function testDuplicateInputVariables() returns error? {
+    string document = string`($userName:string, $userName:int){ greet (name: $userName) }`;
```
ThisaruGuruge, line 23: Shouldn't the type names start with a capital letter? Check the other places as well. If this test is passing, there might be an error in the parsing logic.

ballerina/engine.bal
```diff
@@ -62,16 +62,26 @@ isolated class Engine {
         return getOutputObjectFromErrorDetail(errorDetail);
     }
 
-    isolated function validateDocument(parser:DocumentNode document) returns OutputObject? {
+    isolated function validateDocument(parser:DocumentNode document, map<json>? variables) returns OutputObject? {
+        FragmentVisitor fragmentVisitor = new(document);
```
ThisaruGuruge, line 66: Shall we move the `operationNode.getErrors()` check from `ValidatorVisitor` to here?

ballerina/tests/22_variables.bal
```diff
@@ -0,0 +1,220 @@
...
+
+@test:Config {
+    groups: ["variables", "input"]
+}
+isolated function testInputVariablesWithInvalidArgumentType() returns error? {
+    string document = string`query Greeting($userName:String){ greet (name: $userName ) }`;
+    json variables = { userName: 4 };
+    string url = "http://localhost:9091/inputs";
+    json actualPayload = check getJsonPayloadFromBadRequest(url, document, variables);
+    json expectedPayload = check getJsonContentFromFile("invalid_arguments_type_with_input_variables.json");
```
ThisaruGuruge, line 63: `.json` file name doesn't match with the test.

ballerina/variable_validator.bal
```diff
@@ -0,0 +1,147 @@
...
+        }
+    }
+
+    public isolated function setArgumentValue(Scalar value, parser:ArgumentNode argument, parser:Location location) {
+        if argument.getKind() == parser:T_IDENTIFIER {
+            argument.setValue(value);
+        } else if getTypeNameFromValue(value) == getTypeName(argument) {
+            argument.setValue(value);
+        } else if value is int && getTypeName(argument) == FLOAT {
+            argument.setValue(<float> value);
```
ThisaruGuruge, line 136: Since we are coercing the int value to float here, shall we remove the coercing in the `ValidatorVisitor`?

ballerina/modules/parser/parser.bal
```diff
@@ -205,18 +252,24 @@ public class Parser {
...
-            ArgumentValue value = check getArgumentValue(token);
-
-            ArgumentNode argument = new(name, value, <ArgumentType>token.kind);
-            fieldNode.addArgument(argument);
-            token = check self.peekNextNonSeparatorToken();
+            if (token.kind == T_DOLLAR) {
+                token = check self.readNextNonSeparatorToken();
+                //send the variable name as argument value of the argument node
+                ArgumentValue varName = check getArgumentValue(token); 
+                ArgumentNode argument = new(name, varName, <ArgumentType>token.kind, true);
```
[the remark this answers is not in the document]
DimuthuMadushan: since ```token.kind``` is checked at ```getArgumentValue()```,  we can guarantee that ```token.kind``` is always an ```ArgumentType```.  But i think, it's better to use local variable here.

ballerina/engine.bal
ThisaruGuruge: We have to come up with a way to handle the errors in the operation nodes as well.

ballerina/modules/parser/operation_node.bal
DimuthuMadushan: If so,  to access the above functions, we have to pass the ```OperationNode``` up to ```visitField()``` or get the operation node from ```DocumentNode```. Currently,we are passing ```map<VariableDefinition>``` of ```OperationNode``` up to there. WDYT?

ThisaruGuruge: Sounds right. Let's keep it as it iss then.

ballerina/variable_validator.bal
DimuthuMadushan: Currently, this function is used only for variable values. Parameter arguments coercing is done at the ```ValidatorVisitor```. parameters which have values, won't reach here.

ThisaruGuruge: Right, my bad. But, in that case, we do not need to coerce them here, right?

ballerina/engine.bal
```diff
@@ -62,16 +62,26 @@ isolated class Engine {
...
+        ErrorDetail[]? errors = fragmentVisitor.validate();
+        if (errors is ErrorDetail[]) {
+            return getOutputObjectFromErrorDetail(errors);
+        }
         if (self.maxQueryDepth is int) {
             QueryDepthValidator queryDepthValidator = new QueryDepthValidator(document, <int>self.maxQueryDepth);
```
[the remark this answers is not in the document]
ThisaruGuruge: Please do not mark comments as resolved, without resolving them.

ballerina/modules/parser/tests/parser_tests.bal
```diff
@@ -486,3 +486,31 @@ isolated function testFieldAliasWithArguments() returns error? {
...
 }
+
+@test:Config {
+    groups: ["variables", "parser"]
+}
+isolated function testVariables() returns error? {
+    string document = "query getName($profileId:Int = 3) { profile(id:$profileId) { name } }";
+    Parser parser = new(document);
+    DocumentNode documentNode = check parser.parse();
+    OperationNode operationNode = <OperationNode>documentNode.getOperations()[0];
```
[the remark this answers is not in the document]
ThisaruGuruge: Casting is unnecessary here, right?

ballerina/variable_validator.bal
```diff
@@ -0,0 +1,147 @@
...
+
+    public isolated function init(parser:DocumentNode documentNode, map<json>? variableValues){
+        self.documentNode = documentNode;
+        self.visitedVariableDefinitions = [];
+        self.errors = [];
+        if variableValues == () {
+            self.variables = {};
+        } else {
+            self.variables = <map<json>> variableValues;
+        }
```
ThisaruGuruge, line 35: This should be simple. Is casting needed here? 

```suggestion
        self.variables = variableValues == () ? {} : variableValues;
```

ballerina/variable_validator.bal
```diff
@@ -0,0 +1,147 @@
...
+            return self.errors;
+        }
+    }
+
+    public isolated function visitDocument(parser:DocumentNode documentNode, anydata data = ()) {
+        parser:OperationNode[] operations = documentNode.getOperations();
+        foreach parser:OperationNode operationNode in operations {
+            foreach ErrorDetail errorDetail in operationNode.getErrors() {
+                self.errors.push(errorDetail);
+            }
```
ThisaruGuruge, line 50: Shouldn't this be done at the `visitOperation` function? We visit each operation, and if an operation node contains errors, we push them to `self.errors`?

We can early return from the `visitOperation` function if there are errors present too.

ballerina/variable_validator.bal
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    public isolated function visitOperation(parser:OperationNode operationNode, anydata data = ()) {
+        map<parser:VariableDefinition> variableDefinitions = operationNode.getVaribleDefinitions();
+        foreach parser:Selection selection in operationNode.getSelections() {
+            self.visitSelection(selection, variableDefinitions);
+        }
+        string[] keys = variableDefinitions.keys();
+        foreach string name in keys {
+            if self.visitedVariableDefinitions.indexOf(name) == () {
```
ThisaruGuruge, line 62: Seems like this introduces a bug. When we keep the `visitedVariableDefinitions` in global scope, we have a single array of `visitedVariableDefinitions` for the document. But we define variables in the scope of operation.

In this case, when an operation contains an unused variable, but that same variable is used in another operation, this will not throw an error. Is that the expected behavior? Otherwise, we have to update this.

Consider the following example:

```
query getCity($id: Int, $extra: String) {
    profileExtra(id: $id, extra: $extra) {
        city
    }
}

query getNameById($id: Int, $name: String) {
    profile(id: $id) {
        name
    }
}
```

This should return an error since `getNameById` operation has an unused variable. This might depend on the operation order too.

ballerina/variable_validator.bal
```diff
@@ -0,0 +1,144 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import graphql.parser;
+
+class VariableValidator{
```
ThisaruGuruge, line 19:
```suggestion
class VariableValidator {
```

ballerina/variable_validator.bal
ThisaruGuruge: Let's add a test case for this scenario

ballerina/variable_validator.bal
```diff
@@ -0,0 +1,144 @@
...
+        } else {
+            string message = string`Variable "$${<string> argument.getVariableName()}" got invalid value ${value.toString()};` +
+            string`Expected type ${getTypeName(argument)}. ${getTypeName(argument)} cannot represent value: ${value.toString()}`;
+            self.errors.push(getErrorDetailRecord(message, location)); 
+        }
+    }
+
+    isolated function getErrors() returns ErrorDetail[] {
+        return self.errors;
+    }
```
ThisaruGuruge, line 143: This function is never used?

ballerina/listener_utils.bal
```diff
@@ -22,7 +22,12 @@ isolated function handleGetRequests(Engine engine, http:Request request) returns
     string? query = request.getQueryParamValue(PARAM_QUERY);
     if query is string && query != "" {
         string? operationName = request.getQueryParamValue(PARAM_OPERATION_NAME);
-        return getResponseFromQuery(engine, query, operationName);
+        json? variables = request.getQueryParamValue(PARAM_VARIABLES);
+        if variables is map<json> || variables == () {
+            return getResponseFromQuery(engine, query, operationName, <map<json>?> variables);
+        } else {
+            return createResponse("Invalid format in request parameter: variables", http:STATUS_BAD_REQUEST);
```
ThisaruGuruge, line 29: Shall we have a test case for this scenario as well?

ballerina/listener_utils.bal
```diff
@@ -22,7 +22,12 @@ isolated function handleGetRequests(Engine engine, http:Request request) returns
     string? query = request.getQueryParamValue(PARAM_QUERY);
     if query is string && query != "" {
         string? operationName = request.getQueryParamValue(PARAM_OPERATION_NAME);
-        return getResponseFromQuery(engine, query, operationName);
+        json? variables = request.getQueryParamValue(PARAM_VARIABLES);
+        if variables is map<json> || variables == () {
+            return getResponseFromQuery(engine, query, operationName, <map<json>?> variables);
```
ThisaruGuruge, line 27: We don't need to cast here, right?

```suggestion
            return getResponseFromQuery(engine, query, operationName, variables);
```

ballerina/listener_utils.bal
```diff
@@ -43,15 +48,21 @@ isolated function getResponseFromJsonPayload(Engine engine, http:Request request
     var payload = request.getJsonPayload();
     if payload is json {
         var document = payload.query;
+        var variables = payload.variables;
+        variables = variables is error ? () : variables;
         if document is string && document != "" {
-            return getResponseFromQuery(engine, document, getOperationName(payload));
+            if variables is map<json> || variables is () {
+                return getResponseFromQuery(engine, document, getOperationName(payload), <map<json>?> variables);
```
ThisaruGuruge, line 55: Same as above.

ThisaruGuruge (approved):

ThisaruGuruge (approved):

PR #384, opened 2021-08-20
Add Block Strings Support

DimuthuMadushan:
## Purpose

Fixes: [#1704](https://github.com/ballerina-platform/ballerina-standard-library/issues/1704)

## Examples
In GraphQL documents, when we need multi-line strings, we can use block strings as follow:
```graphql
query sendEmail() {
    sendEmail(message: """
        Hello,
          World!,

        Yours,
            GraphQL
    """)
}
```
## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests

### DIFF
--- a/ballerina/modules/parser/lexer.bal
+++ b/ballerina/modules/parser/lexer.bal
@@ -103,22 +103,62 @@ public class Lexer {
         string word = "";
         Location location = self.currentLocation.clone();
         _ = self.readNextChar(); // Ignore first double quote character
-        while (true) {
+        string value = self.charReader.peek();
+        if value == DOUBLE_QUOTE {
+            value = self.readNextChar();
+            value = self.charReader.peek();
+            if value == DOUBLE_QUOTE {
+                return self.readBlockStringLiteral();
+            } else {
+                return  getToken(word, T_STRING, location);
+            }
+        } else {
+            while true {
+                value = self.charReader.peek();
+                if value is LineTerminator {
+                    string message = "Syntax Error: Unterminated string.";
+                    return error UnterminatedStringError(message, line = self.currentLocation.line,
+                                                         column = self.currentLocation.column);
+                } else if value == DOUBLE_QUOTE && previousChar != BACK_SLASH {
+                    value = self.readNextChar();
+                    break;
+                } else {
+                    value = self.readNextChar();
+                    word += value;
+                }
+                previousChar = value;
+            }
+            return getToken(word, T_STRING, location);
+        }
+    }
+
+    isolated function readBlockStringLiteral() returns Token|SyntaxError {
+        string previousChar = "";
+        string word = "";
+        Location location = self.currentLocation.clone();
+        _ = self.readNextChar(); // Ignore third double quote character
+        while true {
             string value = self.charReader.peek();
-            if (value is LineTerminator) {
-                string message = "Syntax Error: Unterminated string.";
-                return error UnterminatedStringError(message, line = self.currentLocation.line,
-                                                     column = self.currentLocation.column);
-            } else if (value == DOUBLE_QUOTE && previousChar != BACK_SLASH) {
-                value = self.readNextChar();
-                break;
+            if value == DOUBLE_QUOTE {
+                previousChar = self.readNextChar();
+                value = self.charReader.peek();
+                if value == DOUBLE_QUOTE {
+                    previousChar = self.readNextChar();
+                    value = self.charReader.peek();
+                    if value == DOUBLE_QUOTE {
+                        value = self.readNextChar();
+                        break;
+                    }
+                    word += previousChar;
+                }
+                word += previousChar;
             } else {
                 value = self.readNextChar();
                 word += value;
             }
             previousChar = value;
         }
-        return getToken(word, T_STRING, location);
+        return getToken(word.trim(), T_STRING, location);
     }
 
     isolated function readNumericLiteral(string fisrtChar) returns Token|SyntaxError {
--- a/ballerina/tests/24_block_strings.bal
+++ b/ballerina/tests/24_block_strings.bal
@@ -1,0 +1,82 @@
+// Copyright (c) 2021, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+//
+// WSO2 Inc. licenses this file to you under the Apache License,
+// Version 2.0 (the "License"); you may not use this file except
+// in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import ballerina/http;
+import ballerina/test;
+
+@test:Config {
+    groups: ["block_strings"]
+}
+isolated function testBlockStrings() returns error? {
+    http:Request request = new;
+    string document = check getGraphQLDocumentFromFile("block_strings.txt");
+    string url = "http://localhost:9091/inputs";
+    json actualPayload = check getJsonPayloadFromService(url, document);
+    json expectedPayload = check getJsonContentFromFile("block_strings.json");
+    test:assertEquals(actualPayload, expectedPayload);
+}
+
+@test:Config {
+    groups: ["block_strings", "variables"]
+}
+isolated function testBlockStringsWithVariables() returns error? {
+    http:Request request = new;
+    string document = check getGraphQLDocumentFromFile("block_strings_with_variables.txt");
+    json variables = { 
+        block: string`
+        This,
+            is
+                Graphql
+                    Block
+                        Strings!` 
+    };
+    string url = "http://localhost:9091/inputs";
+    json actualPayload = check getJsonPayloadFromService(url, document, variables);
+    json expectedPayload = check getJsonContentFromFile("block_strings_with_variables.json");
+    test:assertEquals(actualPayload, expectedPayload);
+}
+
+@test:Config {
+    groups: ["block_strings"]
+}
+isolated function testBlockStringsWithVariableDefaultValue() returns error? {
+    http:Request request = new;
+    string document = check getGraphQLDocumentFromFile("block_string_with_variable_default_value.txt");
+    string url = "http://localhost:9091/inputs";
+    json actualPayload = check getJsonPayloadFromService(url, document);
+    json expectedPayload = check getJsonContentFromFile("block_strings_with_variables_default_value.json");
+    test:assertEquals(actualPayload, expectedPayload);
+}
+
+@test:Config {
+    groups: ["block_strings", "variables"]
+}
+isolated function testBlockStringsWithVariablesIncludedInvalidCharacters() returns error? {
+    http:Request request = new;
+    string document = check getGraphQLDocumentFromFile("block_strings_with_variables.txt");
+    json variables = { 
+        block: string`
+        Hello,
+            World!,
+\"""
+        Yours, """
+            GraphQL` 
+    };
+    string url = "http://localhost:9091/inputs";
+    json actualPayload = check getJsonPayloadFromBadRequest(url, document, variables);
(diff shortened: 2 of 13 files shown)


ballerina/variable_validator.bal
```diff
@@ -128,13 +128,29 @@ class VariableValidator {
         if argument.getKind() == parser:T_IDENTIFIER {
             argument.setValue(value);
         } else if getTypeNameFromValue(value) == getTypeName(argument) {
-            argument.setValue(value);
+            if getTypeNameFromValue(value) == STRING {
+                if self.hasInvalidCharaters(<string>value) {
```
ayeshLK, line 132: Shall we use `value:ensureType(v, string)` to safely cast variables ?

ballerina/variable_validator.bal
```diff
@@ -128,13 +128,29 @@ class VariableValidator {
...
-            self.errors.push(getErrorDetailRecord(message, location)); 
+            self.errors.push(getErrorDetailRecord(message, location));
+        }
+    }
+
+    public isolated function hasInvalidCharaters(string value) returns boolean {
+        if value.includes(string`"""`) || value.includes(string`\"""`)  {
+            return true;
         }
+        return false;
```
ayeshLK, line 154:
```suggestion
    public isolated function hasInvalidCharaters(string value) returns boolean {
        return value.includes(string`"""`) || value.includes(string`\"""`);
```

ballerina/variable_validator.bal
DimuthuMadushan: Since we have checked the value at line 131, Could it be a problem use direct casting here?

ballerina/modules/parser/lexer.bal
```diff
@@ -103,22 +103,62 @@ public class Lexer {
...
-                break;
+            if value == DOUBLE_QUOTE {
+                previousChar = self.readNextChar();
+                value = self.charReader.peek();
+                if value == DOUBLE_QUOTE {
+                    previousChar = self.readNextChar();
+                    value = self.charReader.peek();
+                    if value == DOUBLE_QUOTE {
+                        value = self.readNextChar();
+                        break;
```
ThisaruGuruge, line 150: There are two places where we have to check for the triple-double quotes. Isn't it be easy if we use a separate function for this?

ThisaruGuruge:
There's a test failure in windows build too.

ballerina/modules/parser/lexer.bal
```diff
@@ -103,22 +103,62 @@ public class Lexer {
...
         _ = self.readNextChar(); // Ignore first double quote character
-        while (true) {
+        string value = self.charReader.peek();
+        if value == DOUBLE_QUOTE {
+            value = self.readNextChar();
+            value = self.charReader.peek();
+            if value == DOUBLE_QUOTE {
+                return self.readBlockStringLiteral();
+            } else {
+                return  getToken(word, T_STRING, location);
```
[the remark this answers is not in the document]
ThisaruGuruge: Shall we add a test containing an empty string (`""`) for this too?

ballerina/modules/parser/lexer.bal
```diff
@@ -103,22 +103,112 @@ public class Lexer {
...
+                lines.push(word);
+                word = "";
                 value = self.readNextChar();
-                break;
+            } else if value == DOUBLE_QUOTE {
+                previousChar = self.readNextChar();
+                value = self.charReader.peek();
+                if value == DOUBLE_QUOTE {
+                    previousChar = self.readNextChar();
+                    value = self.charReader.peek();
```
ThisaruGuruge, line 153: Do we have a test for this?
i.e. a case where `""` occurs in a block string? If not shall we add one?

ballerina/modules/parser/lexer.bal
```diff
@@ -103,22 +103,112 @@ public class Lexer {
...
+    isolated function getBlockStringValue(string[] lines) returns string {
+        int? commonIndent = ();
+        string formatted = "";
+        foreach string line in lines {
+            int indent  = self.getLeadingWhiteSpaceCount(line);
+            if indent < line.length() && (commonIndent is () || indent < commonIndent) {
+                commonIndent = indent;
+                if commonIndent == 0 {
+                    break;
+                }
```
ThisaruGuruge, line 181: Is this necessary? This is a special case right? Do we need to consider the special case here, since this will be handled anyways?

ballerina/tests/24_block_strings.bal
```diff
@@ -0,0 +1,66 @@
...
+
+@test:Config {
+    groups: ["block_strings"]
+}
+isolated function testBlockStringsWithVariableDefaultValue() returns error? {
+    http:Request request = new;
+    string document = check getGraphQLDocumentFromFile("block_string_with_variable_default_value.txt");
+    string url = "http://localhost:9091/inputs";
+    json actualPayload = check getJsonPayloadFromService(url, document);
+    json expectedPayload = check getJsonContentFromFile("block_strings_with_variables_default_value.json");
```
ThisaruGuruge, line 40: Name mismatch?
```suggestion
    json expectedPayload = check getJsonContentFromFile("block_string_with_variables_default_value.json");
```

ballerina/tests/resources/documents/block_strings_with_invalid_characters.txt
```diff
@@ -0,0 +1,7 @@
+{
+    sendEmail(message: """Hello,
+            World!,
+\"""
+        Yours,
+            GraphQL""") 
+}
```
ThisaruGuruge, line 7: What is the invalid character in this?

ballerina/tests/resources/expected_results/block_strings_with_invalid_characters.json
```diff
@@ -0,0 +1,13 @@
+{
+    "errors": [
+        {
+            "message": "Syntax Error: Cannot parse the unexpected character \"\"\".",
```
ThisaruGuruge, line 4: Why does this consider an invalid character?

ThisaruGuruge:
There are still test failures in Windows.

ballerina/modules/parser/lexer.bal
DimuthuMadushan: We already have a test named ```testBlockStringsWithDoubleQuotes```.

ballerina/modules/parser/lexer.bal
DimuthuMadushan: will remove it

ballerina/tests/resources/documents/block_strings_with_invalid_characters.txt
DimuthuMadushan: This test name should be ```testInvalidBlockStrings```. I will change it.

ballerina/tests/resources/expected_results/block_strings_with_invalid_characters.json
DimuthuMadushan: This error is not coming from the block string reading. Since this input block string includes a ```""""``` in the middle, rest of the string consider as an identifier. Then this first ```"``` at the end of the the string consider as an invalid character since identifier can't have ```"``` as a character.

DimuthuMadushan: @ThisaruGuruge is this the correct scenario?

ThisaruGuruge: I still don't understand why is this considered an invalid character (or character sequence).

IMHO, any character sequence is accepted within the block string, except the `"""`, which terminates the block.

Even if you try the same block string in the Appollo server, it does not recognize this as an invalid character. What's the basis of considering this an invalid character?

ballerina/modules/parser/lexer.bal
```diff
@@ -103,22 +103,124 @@ public class Lexer {
...
                 value = self.readNextChar();
-                break;
+            } else if value == DOUBLE_QUOTE {
+                string wordWithQuotes = self.readQuotesInBlockStrings(word);
+                if wordWithQuotes == word {
+                    lines.push(word);
+                    break;
+                } else {
+                    word = wordWithQuotes;
+                }
```
ThisaruGuruge, line 155:
```suggestion
                word = wordWithQuotes;
```

We are pushing the line anyways if we find a new line character, right?

ballerina/modules/parser/lexer.bal
```diff
@@ -103,22 +103,124 @@ public class Lexer {
...
                 word += value;
             }
             previousChar = value;
         }
-        return getToken(word, T_STRING, location);
+        return self.getBlockStringValue(lines);
+    }
+
+    isolated function readQuotesInBlockStrings(string word) returns string {
+        string previousChar = self.readNextChar();
```
ThisaruGuruge, line 166: This variable is redundant. It has only been used to concatenate the character inside the loop. The character is not carried forward for the next iteration. We can use a local variable for that.

ballerina/modules/parser/lexer.bal
```diff
@@ -103,22 +103,124 @@ public class Lexer {
...
+        int? commonIndent = ();
+        string formatted = "";
+        foreach string line in lines {
+            int indent  = self.getLeadingWhiteSpaceCount(line);
+            if indent < line.length() && (commonIndent is () || indent < commonIndent) {
+                commonIndent = indent;
+            }
+        }
+        if commonIndent is int {
+            foreach string line in lines {
```
ThisaruGuruge, line 197: Since we need the index of the line inside the loop, better to use a for loop with a number range instead of a foreach loop. 
[Check this BBE](https://ballerina.io/learn/by-example/foreach-statement.html?is_ref_by_example=true)

ballerina/modules/parser/lexer.bal
```diff
@@ -103,22 +103,124 @@ public class Lexer {
...
+                if commonIndent > line.length() {
+                    continue;
+                } else {
+                    lines[<int>lines.indexOf(line)] = line.substring(commonIndent, line.length());
+                }
+            }
+        }
+        foreach string line in lines {
+            formatted += line + "\n";
```
ThisaruGuruge, line 206: Use [join](https://lib.ballerina.io/ballerina/lang.string/1.1.0/functions#'join) instead. 

```suggestion
            formatted += string:'join("\n", lines);
```

ballerina/tests/24_block_strings.bal
ThisaruGuruge: Still, the names does not match :) 

```
block_strings_with_variable_default_value.txt
block_strings_with_variables_default_value.json
```

ballerina/tests/resources/expected_results/invalid_block_strings.json
```diff
@@ -0,0 +1,13 @@
+{
+    "errors": [
+        {
+            "message": "Syntax Error: Cannot parse the unexpected character \"\"\".",
```
ThisaruGuruge, line 4: This error seems to be not right.  In the given document, the block string ends at `{ line: 4, column: 1 }`.
Then there's another identifier, which means we have another argument. Then there's another identifier without a colon character, which is incorrect.

So the error should be 
```
{
    message: "Syntax Error: Expected \":\", found Name.",
    locations: [
        {
            line: 6,
            column: 14
        }
    ]
}
```

ballerina/modules/parser/lexer.bal
```diff
@@ -103,22 +103,124 @@ public class Lexer {
...
-                string message = "Syntax Error: Unterminated string.";
-                return error UnterminatedStringError(message, line = self.currentLocation.line,
-                                                     column = self.currentLocation.column);
-            } else if (value == DOUBLE_QUOTE && previousChar != BACK_SLASH) {
+            if value == NEW_LINE {
+                lines.push(word);
+                word = "";
                 value = self.readNextChar();
-                break;
+            } else if value == DOUBLE_QUOTE {
```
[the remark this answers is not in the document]
DimuthuMadushan: if we use `else if value == DOUBLE_QUOTE && previousChar != BACK_SLASH {
`, we should use `else if value == DOUBLE_QUOTE && previousChar == BACK_SLASH {
` to remove the `\` in this special case `\" " "`. This may repeat the same thing.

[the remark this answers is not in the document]
ThisaruGuruge: No, the point is to identify the end of the block string, which is `value == DOUBLE_QUOTE && previousChar != BACK_SLASH`, and then two consecutive `"` characters. Otherwise, it is just another character.

PR #393, opened 2021-09-04
Fix Not Allowing NON_NULL Types in Variable Definitions

DimuthuMadushan:
## Purpose

Fixes: [#1803](https://github.com/ballerina-platform/ballerina-standard-library/issues/1803)

## Examples
N/A
## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests


ballerina/modules/parser/lexer.bal
```diff
@@ -388,6 +388,8 @@ isolated function getTokenType(string value) returns TokenType {
         return T_INT;
     } else if (value == HASH) {
         return T_COMMENT;
+    } else if (value == EXCLAMATION) {
```
dilanSachi, line 391: We can remove these parenthesis

ballerina/variable_validator.bal
```diff
@@ -154,4 +172,47 @@ class VariableValidator {
...
+            }
+            return wrapperType;
+        } else {
+            return self.getSchemaTypeFromTypeArray(self.schema.types, typeName);
+        }
+    }
+
+    isolated function getSchemaTypeFromTypeArray(__Type[] types, string typeName) returns __Type? {
+        foreach __Type schemaType in types {
+            if (schemaType?.name.toString() == typeName) {
```
dilanSachi, line 212: Here also

ballerina/modules/parser/parser.bal
```diff
@@ -342,6 +342,17 @@ public class Parser {
         return alias;
     }
 
+    isolated function getTypeIdentifierTokenValue(Token token) returns string|Error {
+        string varType = check getIdentifierTokenvalue(token);
+        Token nextToken = check self.peekNextNonSeparatorToken();
+        if nextToken.kind == T_EXCLAMATION {
```
ThisaruGuruge, line 348: What happens when we find square brackets? Don't we need to handle that?
Names with square brackets are considered in the `getOfTypeNameFromTypeName` function, but I think in this state if the document has a square bracket, it will result in a syntax error.

A better way to handle this is to throw an error stating that the lists are not yet supported.

We also have to be careful about not closing square brackets, which should give a syntax error. 
We should also consider the scenarios like `$info: [Information!]!`, where the list and the element type both are non-nullable types and other permutations of these.

ballerina/variable_validator.bal
```diff
@@ -154,4 +172,47 @@ class VariableValidator {
...
+            if ofType is __Type {
+                wrapperType.ofType = ofType;
+            }
+            return wrapperType;
+        } else {
+            return self.getSchemaTypeFromTypeArray(self.schema.types, typeName);
+        }
+    }
+
+    isolated function getSchemaTypeFromTypeArray(__Type[] types, string typeName) returns __Type? {
```
ThisaruGuruge, line 210: We already have `getTypeFromTypeArray` function. Why do we need to duplicate it?

ballerina/variable_validator.bal
```diff
@@ -154,4 +172,47 @@ class VariableValidator {
...
+                string message = string`Variable "${<string>argNode.getVariableName()}" of type "${varDef.kind}" `+
+                string`used in position expecting type "${getTypeNameFromType(inputValue.'type)}".`;
+                self.errors.push(getErrorDetailRecord(message, argNode.getLocation()));
+            }
+        }
+    }
+
+    isolated function getTypeRecordFromTypeName(string typeName) returns __Type? {
+        //TODO: Improve for 'List' type variables
+        if typeName.endsWith("!") {
```
ThisaruGuruge, line 195: This is checked twice. Once here, and once again in `getOfTypeNameFromTypeName`. We can simplify this.

ballerina/variable_validator.bal
```diff
@@ -154,4 +172,47 @@ class VariableValidator {
...
+            if getOfType(inputValue.'type)?.name != getOfType(actualType)?.name {
+                string message = string`Variable "${<string>argNode.getVariableName()}" of type "${varDef.kind}" `+
+                string`used in position expecting type "${getTypeNameFromType(inputValue.'type)}".`;
+                self.errors.push(getErrorDetailRecord(message, argNode.getLocation()));
+            } else if inputValue.'type.kind == NON_NULL && actualType.kind != NON_NULL && varDef?.defaultValue is () {
+                string message = string`Variable "${<string>argNode.getVariableName()}" of type "${varDef.kind}" `+
+                string`used in position expecting type "${getTypeNameFromType(inputValue.'type)}".`;
+                self.errors.push(getErrorDetailRecord(message, argNode.getLocation()));
+            }
+        }
```
ThisaruGuruge, line 190: What happens in the `else` part? Shouldn't we handle it?

- What if the `actualType` returns `()`?
- What if `inputValue` is `()`?

ballerina/variable_validator.bal
```diff
@@ -154,4 +172,47 @@ class VariableValidator {
...
+        __Type? actualType = self.getTypeRecordFromTypeName(varDef.kind);
+        if inputValue is __InputValue && actualType is __Type {
+            if getOfType(inputValue.'type)?.name != getOfType(actualType)?.name {
+                string message = string`Variable "${<string>argNode.getVariableName()}" of type "${varDef.kind}" `+
+                string`used in position expecting type "${getTypeNameFromType(inputValue.'type)}".`;
+                self.errors.push(getErrorDetailRecord(message, argNode.getLocation()));
+            } else if inputValue.'type.kind == NON_NULL && actualType.kind != NON_NULL && varDef?.defaultValue is () {
+                string message = string`Variable "${<string>argNode.getVariableName()}" of type "${varDef.kind}" `+
+                string`used in position expecting type "${getTypeNameFromType(inputValue.'type)}".`;
+                self.errors.push(getErrorDetailRecord(message, argNode.getLocation()));
```
ThisaruGuruge, line 188: Using `getOfType()` here will create bugs. We should check the exact type against each other. **_IIRC, we discussed this offline and since the types should be structurally equal, we agreed to use structural equivalency directly in this case_**. Stripping the types using `getOfType` and checking the names instead of the type value, will waste that.

These complex checks are redundant in that case too.

ballerina/variable_validator.bal
```diff
@@ -130,9 +142,15 @@ class VariableValidator {
                 self.errors.push(getErrorDetailRecord(message, location));
             }
         } else {
+            __InputValue? inputValue = getInputValueFromArray(inputValues, argumentNode.getName());
+            if inputValue is __InputValue {
+                __Type? inputFieldType = getOfType(inputValue.'type);
+                __InputValue[]? inputFieldValues = inputFieldType?.inputFields;
+                inputValues = inputFieldValues is __InputValue[] ? inputFieldValues : [];
+            }
```
ThisaruGuruge, line 150: What's the `else` logic here? What happens when the input value is `()`?

ballerina/variable_validator.bal
DimuthuMadushan: Both cases are already handled by the `validator_visitor`.

ThisaruGuruge: We use `variable_validator` before the `validator_visitor` right? Is that okay? I mean doesn't it cause any issues here?

DimuthuMadushan: As we discussed offline, `else` logic is skipped here since it checks at `validator_visitor`.

ballerina/variable_validator.bal
DimuthuMadushan: This one is also skipped.

ballerina/variable_validator.bal
```diff
@@ -149,9 +168,65 @@ class VariableValidator {
...
+                name: ()
+            };
+            ofTypeName = typeName.substring(1, typeName.length()-1);
+            ofType = self.getTypeRecordFromTypeName(ofTypeName);
+            if ofType is __Type {
+                wrapperType.ofType = ofType;
+            }
+            return wrapperType;
+        } else {
+            ofType = getTypeFromTypeArray(self.schema.types, typeName);
```
ThisaruGuruge, line 225: What happens if the type is not found here?

Let's say we have `Person!` as the type and the `Person` object type is not found. This will return the following type:

```
{
    "name": (),
    "kind": NON_NULL,
    "ofType": () 
}
```

Which is wrong. We should return an error in this case.

ThisaruGuruge (approved):

ballerina/variable_validator.bal
```diff
@@ -69,50 +75,56 @@ class VariableValidator {
...
                 } else if variableDefinition?.defaultValue is parser:ArgumentValue {
                     parser:ArgumentValue value = <parser:ArgumentValue> variableDefinition?.defaultValue;
-                    argumentNode.setKind(getArgumentTypeKind(variableDefinition.kind));
+                    argumentNode.setKind(getArgumentTypeKind(getOfTypeNameFromTypeName(variableDefinition.kind)));
                     argumentNode.setValue(argumentNode.getName(), value);
                     argumentNode.setVariableDefinition(false);
                 } else if variableDefinition?.defaultValue is parser:ArgumentNode {
                     parser:ArgumentNode value = <parser:ArgumentNode> variableDefinition?.defaultValue;
-                    argumentNode.setKind(getArgumentTypeKind(variableDefinition.kind));
+                    argumentNode.setKind(getArgumentTypeKind(getOfTypeNameFromTypeName(variableDefinition.kind)));
```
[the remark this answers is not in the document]
DimuthuMadushan: `ArgumentType` can have one of `T_INT|T_FLOAT|T_BOOLEAN|T_STRING|T_IDENTIFIER`. For an input object, it will be `T_IDENTIFIER`.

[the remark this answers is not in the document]
ThisaruGuruge: It is wrong, right? `T_IDENTIFIER` is used for enum values. That means, a single identifier token. It should not be used to identify input objects, which may cause confusion, and some issue later, too.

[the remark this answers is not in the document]
DimuthuMadushan: May be then we have to change the `ArgumentNode` and parser as well. Shall we track this via separate issue?

ballerina/variable_validator.bal
DimuthuMadushan: Since this is also checked at the `checkVariableUsageCompatibility( )`, error is not added to this function. Will change it.

ballerina/variable_validator.bal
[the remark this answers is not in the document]
ThisaruGuruge: +1

ballerina/variable_validator.bal
ThisaruGuruge: So are we traversing the same variable twice? Can't we handle all this within the same iteration?

As I mentioned in a previous comment, we are duplicating the condition as well right? Better if we can handle it here. (Or handle this there).

ballerina/variable_validator.bal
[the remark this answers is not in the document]
DimuthuMadushan: will create.

ballerina/variable_validator.bal
ThisaruGuruge: What's the resolution for this?

Using the same string operations on the same string twice does not seems good. We can use this function to get the type, and if we want the type name we can use `getOfType()?.name` which is efficient.

ballerina/variable_validator.bal
```diff
@@ -149,9 +168,87 @@ class VariableValidator {
...
+        }
+        return false;
+    }
+
+    isolated function areTypesCompatible(__Type varType, __Type inputType) returns boolean {
+        if inputType.kind == NON_NULL {
+            if varType.kind != NON_NULL {
+                return false;
+            }
+            return self.areTypesCompatible(getOfType(varType), getOfType(inputType));
```
ThisaruGuruge, line 208: `getOfType()` here will strip the type completely, which means removes all the wrapping types. Is that what we need here? Or do we need just `varType.ofType` ?

ballerina/variable_validator.bal
```diff
@@ -149,9 +168,87 @@ class VariableValidator {
...
+        } else if varType.kind == NON_NULL {
+            return self.areTypesCompatible(getOfType(varType), inputType);
```
ThisaruGuruge, line 210: Same as above comment

ballerina/variable_validator.bal
DimuthuMadushan: This should be `varType.ofType`. Here we want to remove only the first wrapper type.

ballerina/variable_validator.bal
DimuthuMadushan: Then we may have to use string casting there since `getOfType()?.name` is optional field and bring checks in `checkVariableUsageCompatibility` into `visitArgument` since we don't have exact input type in `visitArgument`.

ballerina/modules/parser/tests/parser_tests.bal
```diff
@@ -509,6 +509,94 @@ isolated function testVariables() returns error? {
...
+    groups: ["variables", "list", "parser"]
+}
+isolated function testInvalidListTypeVariableMissingCloseBracket() returns error? {
+    string document = "query getId($name: [String![) { profile(userName:$name) { id } }";
+    Parser parser = new(document);
+    DocumentNode|Error result = parser.parse();
+    test:assertTrue(result is InvalidTokenError);
+    InvalidTokenError err = <InvalidTokenError>result;
+    string expectedMessage = string`Syntax Error: Expected "]", found "[".`;
+    test:assertEquals(err.message(), expectedMessage);
```
ThisaruGuruge, line 584: Shouldn't we assert the error location as well? As a practice, when there's an error, let's check the error location as well.

ThisaruGuruge (approved):

PR #417, opened 2021-09-23
Add Context Support for GraphQL Resolvers

ThisaruGuruge:
## Purpose
Add GraphQL context support that can be used to share metadata between the resolvers.

Fixes: [#1906](https://github.com/ballerina-platform/ballerina-standard-library/issues/1906)
Fixes: [#1953](https://github.com/ballerina-platform/ballerina-standard-library/issues/1953)

## Examples

Developers can provide a function to initialize the context which must be an isolated function, and must have two input parameters: `http:Request` and `http:RequestContext`. This function must have the return type of `graphql:Context|error`. 

Inside this function, the `graphql:Context` object can be created and the developer can add attributes to the context as key-value pairs. The key is always a string, the value can be a `value:Clonable|isolated object {}`. Simply put, this can be a primitive value, any `readonly` value, or an isolated function pointer, or an isolated object. The context object has three functions, `add`, `get`, and `remove`. 
Then this function can be provided as a service config field. Using this function, a context will be created per request. When needed, the context can be defined as the **first parameter** of a resolver (resource/remote) function. Following is an example:

```ballerina
import ballerina/graphql;
import ballerina/http;

isolated function initContext(http:Request request, http:RequestContext requestContext) returns graphql:Context|error {
    graphql:Context context = new;
    check context.add("scope", request.getHeader("scope"));
    return context;
}

@graphql:ServiceConfig {
    contextInit: initContext
} 
service on new graphql:Listener(9000) {
    resource function get greting(graphql:Context context) returns string {
        if context.get("scope") is string && context.get("scope") == "admin" {
            return "Hello, admin!";
        } else {
            return "Hello, user!";
        }
    }
}
```

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests

### DIFF
--- a/ballerina/annotations.bal
+++ b/ballerina/annotations.bal
@@ -14,8 +14,6 @@
 // specific language governing permissions and limitations
 // under the License.
 
-import ballerina/http;
-
 # Provides a set of configurations for the GraphQL service.
 #
 # + maxQueryDepth - The maximum depth allowed for a query
@@ -24,10 +22,8 @@ import ballerina/http;
 public type GraphqlServiceConfig record {|
     int maxQueryDepth?;
     ListenerAuthConfig[] auth?;
-    ContextInit contextInit?;
+    ContextInit contextInit = initDefaultContext;
 |};
 
 # The annotation to configure a GraphQL service.
 public annotation GraphqlServiceConfig ServiceConfig on service;
-
-type ContextInit isolated function (http:Request request, http:RequestContext requestContext) returns Context|error;
--- a/ballerina/annotations.bal
+++ b/ballerina/annotations.bal
@@ -14,14 +14,20 @@
 // specific language governing permissions and limitations
 // under the License.
 
+import ballerina/http;
+
 # Provides a set of configurations for the GraphQL service.
 #
 # + maxQueryDepth - The maximum depth allowed for a query
 # + auth - Listener authenticaton configurations
+# + contextInit - Function to initialize the context. If not provided, an empty context will be created
 public type GraphqlServiceConfig record {|
     int maxQueryDepth?;
     ListenerAuthConfig[] auth?;
+    ContextInit contextInit?;
 |};
 
 # The annotation to configure a GraphQL service.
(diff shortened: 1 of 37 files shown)


ballerina/annotations.bal
```diff
@@ -14,14 +14,20 @@
...
+
 # Provides a set of configurations for the GraphQL service.
 #
 # + maxQueryDepth - The maximum depth allowed for a query
 # + auth - Listener authenticaton configurations
+# + contextInit - Function to initialize the context. If not provided, an empty context will be created
 public type GraphqlServiceConfig record {|
     int maxQueryDepth?;
     ListenerAuthConfig[] auth?;
+    ContextInit contextInit?;
```
shafreenAnfar, line 27: What if we assign a default value for this ? Something like

```ballerina
isolated function initDefaultContext(http:Request request, http:RequestContext requestContext) returns graphql:Context|error {
    return graphql:Context();
}
```

That way it is explicit that we create a default one if nothing is given.

ThisaruGuruge: We already have a function for this:
https://github.com/ballerina-platform/module-ballerina-graphql/blob/df9a8fae41edc30c81f6616044378b866008a943/ballerina/context.bal#L68

ldclakmal: @ThisaruGuruge @shafreenAnfar What do you think about the field name `contextInit`? I prefer this to be as `Context context`, and passing a function pointer which creates a `Context` for me. But I understand that it is not possible with Ballerina way of handling types. So, what do you think of having this as `ContextInit context`?

ThisaruGuruge: At first, I went through the same thought process. But then went with `contextInit` because this isn't actually a `context`, but a function that initializes a context.
IMHO, this makes it easy for the developers to understand what it is. They have to provide a `ContextInit` (which is a function pointer), not a `Context` (which is an object).

ldclakmal: Yeah. Agree with the clarification.

shafreenAnfar (approved):

PR #476, opened 2021-11-05
Skip Master Build for `.md` File Changes

ThisaruGuruge:
## Purpose
$Subject

Related: [#2362](https://github.com/ballerina-platform/ballerina-standard-library/issues/2362)

## Examples
N/A

## Checklist
- [x] Linked to an issue
- [ ] ~Updated the changelog~
- [ ] ~Added tests~


ballerina/Dependencies.toml
```diff
@@ -195,9 +195,7 @@ org = "ballerina"
 name = "lang.runtime"
 version = "0.0.0"
 dependencies = [
-	{org = "ballerina", name = "jballerina.java"},
-	{org = "ballerina", name = "lang.array"},
-	{org = "ballerina", name = "lang.value"}
+	{org = "ballerina", name = "jballerina.java"}
```
niveathika, line 198: A tangent issue, is this change happening only for `-x test` build? If so we need to run this commit task only is tests are enabled or not?

.github/workflows/build-timestamped-master.yml
```diff
@@ -4,6 +4,9 @@ on:
     push:
         branches:
             - master
+        paths-ignore:
+            - '**/.md'
```
niveathika, line 8: This will ignore Module.md changes too, won't it? Would be an issue if we have to check any doc changes

ballerina/Dependencies.toml
ThisaruGuruge: Nope, I did a full build.

.github/workflows/build-timestamped-master.yml
ThisaruGuruge: Good point. We may have to include them separately. Will update the PR.

ballerina/Dependencies.toml
niveathika: I noticed this behaviour when building without tests, the dependencies get updated and reverts again. This is adding unnecessary automatic commits.

.github/workflows/build-timestamped-master.yml
ThisaruGuruge: Fixed with d881a21021909eb442dc765d441ba5be7c0bccca

ballerina/Dependencies.toml
ThisaruGuruge: Cool.

niveathika (approved):

PR #484, opened 2021-11-11
Fix service detaching logic in listener

ThisaruGuruge:
## Purpose
Previously, when the `detach` function of the GraphQL listener is called, no matter which service is passed to detach, the latest attached HTTP service is being detached from the underlying HTTP listener. This causes unexpected results.
This PR will fix this issue, by keeping track the HTTP service related to a particular GraphQL service. 

Fixes: [#2356](https://github.com/ballerina-platform/ballerina-standard-library/issues/2356)

## Examples
N/A

## Checklist
- [X] Linked to an issue
- [X] Updated the changelog
- [X] Added tests

### DIFF
--- a/ballerina/listener.bal
+++ b/ballerina/listener.bal
@@ -19,7 +19,6 @@ import ballerina/http;
 # Represents a Graphql listener endpoint.
 public class Listener {
     private http:Listener httpListener;
-    private HttpService? httpService;
 
     # Invoked during the initialization of a `graphql:Listener`. Either an `http:Listner` or a port number must be
     # provided to initialize the listener.
@@ -39,7 +38,6 @@ public class Listener {
         } else {
             self.httpListener = listenTo;
         }
-        self.httpService = ();
     }
 
     # Attaches the provided service to the Listener.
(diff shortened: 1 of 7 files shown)


ballerina/listener.bal
```diff
@@ -57,20 +55,22 @@ public class Listener {
...
 
     # Detaches the provided service from the Listener.
     #
     # + s - The service to be detached from the listener
     # + return - A `graphql:Error` if an error occurred during the service detaching process or else `()`
     public isolated function detach(Service s) returns Error? {
-        if (self.httpService is HttpService) {
-            error? result = self.httpListener.detach(<HttpService>self.httpService);
+        HttpService? httpService = getHttpServiceFromGraphqlService(s);
+        if (httpService is HttpService) {
```
ayeshLK, line 72:
```suggestion
        if httpService is HttpService {
```

ThisaruGuruge: Fixed with aa023667cdecea03ed6b031c4bff96e16f7e1ebc

native/src/main/java/io/ballerina/stdlib/graphql/runtime/engine/ListenerUtils.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package io.ballerina.stdlib.graphql.runtime.engine;
+
+import io.ballerina.runtime.api.values.BObject;
+
+/**
+ * External utility methods used in Ballerina GraphQL listener.
+ */
+public class ListenerUtils {
```
ayeshLK, line 26: Since, this is a utility class we could mark it as `final` (since we do not need developers to create sub-classes out of this).
```suggestion
public final class ListenerUtils {
```

DimuthuMadushan (approved):

PR #489, opened 2021-11-15
Migrate to latests ballerina runtime APIs

DimuthuMadushan:
Related to: [#2383](https://github.com/ballerina-platform/ballerina-standard-library/issues/2383) 

## Checklist
- [x] Linked to an issue
- [ ] Updated the changelog
- [ ] Added tests

### DIFF
--- a/native/src/main/java/io/ballerina/stdlib/graphql/runtime/engine/Engine.java
+++ b/native/src/main/java/io/ballerina/stdlib/graphql/runtime/engine/Engine.java
@@ -168,7 +168,7 @@ public class Engine {
         ResourceCallback callback =
                 new ResourceCallback(executionContext, node, data, pathSegments);
         executionContext.getCallbackHandler().addCallback(callback);
-        if (service.getType().isIsolated()) {
+        if (service.getType().isIsolated() && service.getType().isIsolated(method.getName())) {
             executionContext.getEnvironment().getRuntime()
                     .invokeMethodAsyncConcurrently(service, method.getName(), null,
                             strandMetadata, callback, null, PredefinedTypes.TYPE_NULL, args);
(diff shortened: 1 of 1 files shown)


shafreenAnfar (dismissed):

native/src/main/java/io/ballerina/stdlib/graphql/runtime/engine/Engine.java
```diff
@@ -167,8 +168,15 @@ private static void getExecutionResult(ExecutionContext executionContext, BObjec
         ResourceCallback callback =
                 new ResourceCallback(executionContext, node, data, pathSegments);
         executionContext.getCallbackHandler().addCallback(callback);
-        executionContext.getEnvironment().getRuntime().invokeMethodAsync(service, method.getName(), null,
-                                                                         strandMetadata, callback, args);
+        if (service.getType().isIsolated()) {
+            executionContext.getEnvironment().getRuntime()
+                    .invokeMethodAsyncConcurrently(service, method.getName(), null,
+                            strandMetadata, callback, null, PredefinedTypes.TYPE_NULL, args);
```
ayeshLK, line 174:
```suggestion
        if (service.getType().isIsolated(method.getName())) {
            executionContext.getEnvironment().getRuntime()
                    .invokeMethodAsyncConcurrently(service, method.getName(), null,
                            strandMetadata, callback, null, PredefinedTypes.TYPE_NULL, args);
```

shafreenAnfar: I was under the impression that `service.getType().isIsolated()` returns `true` if the service instance is isolated. If that is the case why do we need to check if a given method is isolated ?

ayeshLK: @shafreenAnfar if the service-object is isolated it does not guarantee that the methods in that service-object is also isolated. It guarantees that the fields inside that service is safe to use in isolation. But if we want to run a service-method concurrently service being isolated is not enough. The method itself should be isolated. That is why we need to use `service.getType().isIsolated(methodName)`. Internally this checks whether service-object is also isolated and then check whether method is isolated.

ayeshLK (approved):

PR #497, opened 2021-11-19
Add Input Value List Support for GraphQL 

DimuthuMadushan:
## Purpose

Fixes: [#1837](https://github.com/ballerina-platform/ballerina-standard-library/issues/1837)
Fixes: [#2133](https://github.com/ballerina-platform/ballerina-standard-library/issues/2133)

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests


ballerina/validator_visitor.bal
```diff
@@ -227,12 +294,78 @@ class ValidatorVisitor {
...
+                foreach int i in 0..< variableValues.length() {
+                    self.updatePath(i);
+                    json listItemValue = variableValues[i];
+                    if listItemValue is Scalar {
+                        parser:ArgumentValue argValue = listItemValue;
+                        if getOfType(listItemInputValue.'type).kind == ENUM {
+                            self.validateEnumArgument(argValue, location, ENUM, listItemInputValue);
+                        } else {
+                            string expectedTypeName = getOfType(listItemInputValue.'type).name.toString();
+                            if expectedTypeName == FLOAT && argValue is int {
```
ThisaruGuruge, line 318: Isn't the `coerceArgumentIntValueToFloat` function does the same (or something similar)?

changelog.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Added
 - [[#741] Add Default Directives 'skip' & 'include' Support](https://github.com/ballerina-platform/ballerina-standard-library/issues/741)
 - [[#2026] Implement auth error types](https://github.com/ballerina-platform/ballerina-standard-library/issues/2026)
+- [[#1837] Add Input Value List Support](https://github.com/ballerina-platform/ballerina-standard-library/issues/1837)
```
ThisaruGuruge, line 12: Since we have already released the `1.0.1`, we have to move this under the new release.

compiler-plugin-tests/src/test/resources/ballerina_sources/invalid_service_26/service.bal
```diff
@@ -0,0 +1,76 @@
...
+    Person author;
+};
+
+type Location record {
+    float latitude;
+    float longitude;
+};
+
+service /graphql on new graphql:Listener(4000) {
+    resource function get profile(Person[] p) returns Person {
```
ThisaruGuruge, line 35: Shall we add another test for the following case?

```ballerina
resource function get profile(Person[] p) returns Person[] {
}
```

native/src/main/java/io/ballerina/stdlib/graphql/runtime/engine/Engine.java
```diff
@@ -179,67 +188,169 @@ private static void getExecutionResult(ExecutionContext executionContext, BObjec
...
+            argumentsMap.put(argName, value);
+        }
+    }
+
+    public static void addListMembersFromVariableValue(BObject listItem, BArray valueArray, Type argType) {
+        if (argType.getTag() == TypeTags.RECORD_TYPE_TAG) {
+            BMap<BString, Object> inputObjectFields = listItem.getMapValue(VARIABLE_VALUE_FIELD);
+            BMap<BString, Object> inputObjectRecord = ValueCreator.createMapValue(argType);
+            visitInputObjectTypeVariableValue(inputObjectFields, inputObjectRecord, (RecordType) argType);
+            valueArray.append(inputObjectRecord);
```
ThisaruGuruge, line 296: Why does this line not covered in tests? If a test is missing, shall we add one?

native/src/main/java/io/ballerina/stdlib/graphql/runtime/utils/Utils.java
```diff
@@ -66,6 +67,9 @@ public static BError createError(String message, String errorTypeName, BError ca
     }
 
     public static boolean isContext(Type type) {
+        if (type.getTag() == TypeTags.ARRAY_TAG) {
+            return false;
+        }
```
ThisaruGuruge, line 72: Why do we have to specifically check for the array types here? Is this check needed?

spotbugs-exclude.xml
```diff
@@ -6,4 +6,20 @@
...
+            </And>
+            <Bug pattern="BC_UNCONFIRMED_CAST" />
+        </And>
+    </Match>
+    <Match>
+        <And>
+            <Class name="io.ballerina.stdlib.graphql.runtime.engine.Engine" />
+            <Method name="addInputObjectFieldsFromVariableValue" />
+            <Bug pattern="BC_UNCONFIRMED_CAST" />
+        </And>
```
ThisaruGuruge, line 23: We can combine these two by removing the method name.

ballerina/validator_visitor.bal
DimuthuMadushan: `coerceArgumentIntValueToFloat` function is used for argument nodes. Since this is a variable value, here we get a `json[]`. Shall we have a separate function? This is the only place we get a `json[]` member to coerce.

native/src/main/java/io/ballerina/stdlib/graphql/runtime/engine/Engine.java
DimuthuMadushan: This test is missing. Will add one.

native/src/main/java/io/ballerina/stdlib/graphql/runtime/utils/Utils.java
DimuthuMadushan: When we get an array type input like `float[][]`, `type.Package` value is not present. Due to that following if condition will cause to a NPE. `if (type.getPackage().getOrg() == null || type.getPackage().getName() == null) {`.

spotbugs-exclude.xml
DimuthuMadushan: if we remove the method name, Isn't it excluding the unconfirmed cast in the whole 'Engine' class?

ThisaruGuruge: That would be okay I think since we have similar castings all over this class. (This actually is a false positive, as we check the type tag before casting, but the spotbugs plugin expects us to use the `instanceof` check.)

ballerina/modules/parser/argument_node.bal
```diff
@@ -19,16 +19,18 @@ public class ArgumentNode {
 
     private string name;
     private Location location;
-    private map<ArgumentValue|ArgumentNode> value;
+    private ArgumentValue|ArgumentValue[] value;
+    private Location valueLocation;
     private ArgumentType kind;
     private string? variableName;
-    private anydata? variableValue;
+    private anydata variableValue;
```
ThisaruGuruge, line 26: Can't this be nil?

ballerina/validator_visitor.bal
ThisaruGuruge: Right. In that case, this will be okay.

native/src/main/java/io/ballerina/stdlib/graphql/runtime/utils/Utils.java
ThisaruGuruge: Then let's add a check like this:

```java
if (type.getPackage() == null) {
    return false;
}

if (type.getPackage().getOrg() == null || type.getPackage().getName() == null) {
    return false;
}
```

ThisaruGuruge (approved):

PR #530, opened 2021-12-20
Add GraphQL Load Tests

ThisaruGuruge:
## Purpose
$Subject

Fixes: [#2242](https://github.com/ballerina-platform/ballerina-standard-library/issues/2242)

## Examples
N/A

## Checklist
- [x] Linked to an issue
- [ ] ~Updated the changelog~
- [ ] ~Added tests~


.github/workflows/trigger-load-tests.yml
```diff
@@ -0,0 +1,48 @@
+name: Trigger Load Tests
+
+on:
+    workflow_dispatch:
+        inputs:
+            tests:
+                description: >
+                    List of test names. This needs to be filled only if you want to run a specific set of tests. Example: foo,bar
+                required: false
+
```
dilanSachi, line 10: Better to add a cluster name input also. when we want to run the load test before another test finishes, we need to change the cluster name because the test framework sometimes fails to create the clusters when another cluster is there with the same name

ThisaruGuruge: Fixed with ba7609d608087066771242721fd2a3fa956b546e

.github/workflows/process-load-test-result.yml
```diff
@@ -0,0 +1,55 @@
+name: Process load test results
+on:
+    repository_dispatch:
+        types: [ graphql-load-test ]
+
+jobs:
+    process-results:
+        runs-on: ubuntu-latest
+        steps:
+            -   name: Checkout repository
```
dilanSachi, line 10: is the indentation correct?

ThisaruGuruge: Yes.

dilanSachi (approved):
LGTM

PR #537, opened 2021-12-21
Add CORS configure support for GraphQL services

DimuthuMadushan:
## Purpose

Fixes: [#1475](https://github.com/ballerina-platform/ballerina-standard-library/issues/1475)

## Examples
GraphQL `@ServiceConfig` annotation can be used to configure the CORS for GraphQL services as following example:
```ballerina
@ServiceConfig {
    cors: {
        allowOrigins : ["http://www.wso2.com"],
        allowCredentials : true,
        allowHeaders: ["X-Content-Type-Options", "X-PINGOTHER"],
        exposeHeaders: ["X-HEADER"],
        allowMethods : ["POST"]
    }
}

service /graphql on new graphql:Listener(4000) {
    isolated resource function get greet() returns string {
        return "Hello";
    }
}
```

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests


ballerina/annotation_processor.bal
```diff
@@ -14,6 +14,13 @@
 // specific language governing permissions and limitations
 // under the License.
 
+isolated function getCorsConfig(GraphqlServiceConfig? serviceConfig) returns CorsConfig {
+    if serviceConfig is GraphqlServiceConfig && serviceConfig.cors is CorsConfig {
+        return <CorsConfig> serviceConfig.cors;
+    }
+    return {};
```
ThisaruGuruge, line 21: We can return `CorsConfig?` here?

ballerina/listener_utils.bal
```diff
@@ -285,6 +285,64 @@ isolated function forwardMultipartRequestToExecution(map<Upload|Upload[]> fileIn
...
 }
 
+isolated function getHttpService(Engine gqlEngine, GraphqlServiceConfig? serviceConfig) returns HttpService {
+    final readonly & ListenerAuthConfig[]? authConfigurations = getListenerAuthConfig(serviceConfig).cloneReadOnly();
+    final ContextInit contextInitFunction = getContextInit(serviceConfig);
+    final CorsConfig corsConfig = getCorsConfig(serviceConfig);
+
+    HttpService httpService = @http:ServiceConfig {
+        cors: corsConfig
+    } isolated service object {
```
ThisaruGuruge, line 295: Can't we add other configs such as auth configs here as well? @ldclakmal ?

ldclakmal: Yes, we should be able to. That way we can remove the complete logic of auth handler creation from GraphQL. But, the only concern I have with adding auth configs here is that the response message in case of auth failure. Because HTTP responds as 401/403 responses, and that is not applicable in GraphQL context. We may have to add interceptor to modify that response if we are doing. So, shall we keep that as it is and check for a better approach if we have?

ballerina/annotation_processor.bal
DimuthuMadushan: As per the api docs, the default value of HTTP cors config is `{}`. Shall we keep this as it is?

ballerina/listener_utils.bal
ThisaruGuruge: +1
Let's go ahead with this approach for now.

ThisaruGuruge (approved):

shafreenAnfar (approved):

PR #595, opened 2022-02-03
Fix Inter-Dependent Fragments Returning Stack Overflow Error

DimuthuMadushan:
## Purpose

Fixes: [#2640](https://github.com/ballerina-platform/ballerina-standard-library/issues/2640)

## Examples
N/A

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests

### DIFF
--- a/ballerina/common_utils.bal
+++ b/ballerina/common_utils.bal
@@ -103,7 +103,7 @@ isolated function getCycleRecursiveFragmentError(parser:FragmentNode spread, map
         }
         spreadPath += ",";
     }
-    string message = string `Cannot spread fragment "${spread.getName()}" within itself via ${spreadPath}.`;
+    string message = string `Cannot spread fragment "${spread.getName()}" within itself via${spreadPath}.`;
     return getErrorDetailRecord(message, locations);
 }
 
--- a/ballerina/common_utils.bal
+++ b/ballerina/common_utils.bal
@@ -83,6 +83,22 @@ isolated function getInvalidArgumentValueError(string listIndexOfError, string e
     }
 }
 
+isolated function getCycleRecursiveFragmentError(string spreadName, map<parser:FragmentNode> visitedSpreads)
+    returns ErrorDetail {
+    Location[] locations = [];
+    string[] keys = visitedSpreads.keys();
+    int listIndex = <int>keys.indexOf(spreadName);
+    foreach int i in listIndex ..< keys.length() {
+        parser:FragmentNode fragmentSpread = visitedSpreads.get(keys[i]);
+        locations.push(<Location>fragmentSpread.getSpreadLocation());
+    }
+    string message = string`Cannot spread fragment "${spreadName}" within itself.`;
+    if locations.length() > 1 {
+        message = string`Cannot spread fragment "${spreadName}" within itself via "${keys[listIndex + 1]}".`;
+    }
+    return getErrorDetailRecord(message, locations);
+}
+
 isolated function getOutputObject(Data data, ErrorDetail[] errors) returns OutputObject {
     OutputObject outputObject = {};
     if data.length() > 0 {
--- a/ballerina/common_utils.bal
+++ b/ballerina/common_utils.bal
@@ -83,19 +83,27 @@ isolated function getInvalidArgumentValueError(string listIndexOfError, string e
     }
 }
 
-isolated function getCycleRecursiveFragmentError(string spreadName, map<parser:FragmentNode> visitedSpreads)
+isolated function getCycleRecursiveFragmentError(parser:FragmentNode spread, map<parser:FragmentNode> visitedSpreads)
     returns ErrorDetail {
     Location[] locations = [];
-    string[] keys = visitedSpreads.keys();
-    int listIndex = <int>keys.indexOf(spreadName);
-    foreach int i in listIndex ..< keys.length() {
-        parser:FragmentNode fragmentSpread = visitedSpreads.get(keys[i]);
+    string spreadPath = "";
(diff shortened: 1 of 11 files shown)


ballerina/common_utils.bal
```diff
@@ -83,6 +83,22 @@ isolated function getInvalidArgumentValueError(string listIndexOfError, string e
     }
 }
 
+isolated function getCycleRecursiveFragmentError(string spreadName, map<parser:FragmentNode> visitedSpreads)
+    returns ErrorDetail {
+    Location[] locations = [];
+    string[] keys = visitedSpreads.keys();
+    int listIndex = <int>keys.indexOf(spreadName);
+    foreach int i in listIndex ..< keys.length() {
+        parser:FragmentNode fragmentSpread = visitedSpreads.get(keys[i]);
```
ThisaruGuruge, line 92: Can't we replace this by this? We can remove the line no. 88 too.
```suggestion
    foreach string key in visitedSpreads.keys() {
        parser:FragmentNode fragmentSpread = visitedSpreads.get(key);
```

ballerina/common_utils.bal
```diff
@@ -83,6 +83,22 @@ isolated function getInvalidArgumentValueError(string listIndexOfError, string e
     }
 }
 
+isolated function getCycleRecursiveFragmentError(string spreadName, map<parser:FragmentNode> visitedSpreads)
+    returns ErrorDetail {
+    Location[] locations = [];
+    string[] keys = visitedSpreads.keys();
+    int listIndex = <int>keys.indexOf(spreadName);
```
ayeshLK, line 90: Shall we use `value:ensureType` rather than casting ?

ballerina/fragment_visitor.bal
```diff
@@ -20,16 +20,26 @@ class FragmentVisitor {
...
+            if !fragmentNode.isInlineFragment() && self.visitedFragments.indexOf(fragmentNode.getName()) is () {
+                self.detectCycles(fragmentNode, {});
+            }
+        }
+        if self.errors.length() > 0 {
+            return self.errors;
+        }
         self.visitDocument(self.documentNode);
         if self.errors.length() > 0 {
             return self.errors;
```
ThisaruGuruge, line 45: Is there a way to remove this duplicated code? I understand why the same check has to be done twice, but can we remove the duplication somehow?

ballerina/tests/14_fragments.bal
```diff
@@ -289,3 +289,61 @@ isolated function testNestedFragmentsQueryingServiceObjectsWithMultipleFields()
...
+                    {
+                        line: 21,
+                        column: 9
+                    },
+                    {
+                        line: 26,
+                        column: 5
+                    }
+                ]
+            }
```
ThisaruGuruge, line 322: Shall we move this to a separate file?

ballerina/common_utils.bal
DimuthuMadushan: We don't need to iterate the entire array here. Maybe we can slice the array.

ThisaruGuruge: Hmm. In that case, this is okay. I misread the code it seems.

ballerina/common_utils.bal
DimuthuMadushan: Since we already know that `spreadName` is included in the `keys` array, is that ok to directly casting here? Otherwise `value:ensureType` returns `int|error` and we need to handle the error explicitly. don't we? WDYT @ThisaruGuruge

ThisaruGuruge: > Since we already know that `spreadName` is included in the `keys` array, is that ok to directly casting here? Otherwise `value:ensureType` returns `int|error` and we need to handle the error explicitly. don't we? WDYT @ThisaruGuruge

+1 to cast.

ballerina/common_utils.bal
```diff
@@ -83,6 +83,30 @@ isolated function getInvalidArgumentValueError(string listIndexOfError, string e
...
+    returns ErrorDetail {
+    Location[] locations = [];
+    string spreadPath = "";
+    int listIndex = <int>visitedSpreads.keys().indexOf(spread.getName());
+    string[] spreadPathArray = visitedSpreads.keys().slice(listIndex);
+    if spreadPathArray.length() == 1 {
+        string message = string `Cannot spread fragment "${spread.getName()}" within itself.`;
+        return getErrorDetailRecord(message, <Location>spread.getSpreadLocation());
+    }
+    foreach int i in 1 ..< spreadPathArray.length() {
```
ThisaruGuruge, line 96: What will happen if multiple recursive references are there inside the same fragment?

Eg.:

```graphql
{
    profile {
        ...PersonFragment
    }
}

fragment PersonFragment {
    ...Fragment1
    ...Fragment2
}

fragment Fragment1 {
    name
    ...PersonFragment
}

fragment Fragment2 {
    age
    ...PersonFragment
}
```

Shall we add such test as well?

ballerina/tests/resources/expected_results/fragments_with_cycles.json
```diff
@@ -0,0 +1,25 @@
+{
+    "errors": [
+        {
+            "message": "Cannot spread fragment \"TypeFragment\" within itself via  \"FieldFragment\", \"TypeFragment2\", \"FieldFragment2\".",
```
ThisaruGuruge, line 4: Extra space before `\"FieldFragment\"`

ballerina/common_utils.bal
DimuthuMadushan: added a test

ThisaruGuruge (approved):

PR #616, opened 2022-02-21
Fix Not Allowing to Pass `null` as Variable Input Value

DimuthuMadushan:
## Purpose

Fixes: [#2696](https://github.com/ballerina-platform/ballerina-standard-library/issues/2696)

## Examples
N/A

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests

### DIFF
--- a/ballerina/common_utils.bal
+++ b/ballerina/common_utils.bal
@@ -166,6 +166,15 @@ isolated function getOfType(__Type schemaType) returns __Type {
     }
 }
 
+isolated function getOfTypeName(__Type schemaType) returns string {
+    __Type? ofType = schemaType?.ofType;
+    if ofType is () {
+        return <string>schemaType.name;
+    } else {
+        return <string>getOfType(ofType).name;
+    }
+}
+
 isolated function getTypeNameFromType(__Type schemaType) returns string {
     if schemaType.kind == NON_NULL {
         return string`${getTypeNameFromType(<__Type>schemaType?.ofType)}!`;
--- a/ballerina/common_utils.bal
+++ b/ballerina/common_utils.bal
@@ -167,12 +167,7 @@ isolated function getOfType(__Type schemaType) returns __Type {
 }
 
 isolated function getOfTypeName(__Type schemaType) returns string {
-    __Type? ofType = schemaType?.ofType;
-    if ofType is () {
-        return <string>schemaType.name;
-    } else {
-        return <string>getOfType(ofType).name;
-    }
+    return <string>getOfType(schemaType).name;
 }
 
 isolated function getTypeNameFromType(__Type schemaType) returns string {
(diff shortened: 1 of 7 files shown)


ballerina/variable_validator.bal
```diff
@@ -242,21 +241,23 @@ class VariableValidator {
...
-    public isolated function setArgumentValue(json value, parser:ArgumentNode argument,
-                                              parser:VariableDefinitionNode varDef) {
-        if argument.getKind() == parser:T_INPUT_OBJECT {
-            argument.setVariableValue(value);
-        } else if argument.getKind() == parser:T_LIST {
-            argument.setVariableValue(value);
-        } else if argument.getKind() == parser:T_IDENTIFIER {
+    isolated function setArgumentValue(json value, parser:ArgumentNode argument, string variableTypeName,
+                                       __Type variableType) {
+        if getOfType(variableType).name.toString() == UPLOAD {
```
ThisaruGuruge, line 246: Why do we need to use `toString()` here? Shall we check the usage of `toString()` and try to remove/reduce the usage of `toString()`?

ballerina/validator_visitor.bal
```diff
@@ -345,6 +351,9 @@ class ValidatorVisitor {
                         self.updatePath(listItemInputValue.name);
                         self.validateListVariableValue(listItemValue, listItemInputValue, location, fieldName);
                         self.removePath();
+                    } else if listItemValue is () {
+                        string expectedTypeName = getOfType(listItemInputValue.'type).name.toString();
```
ThisaruGuruge, line 355: Shall we remove the `toString()` here as well? We can use the `string?` as the type here.

DimuthuMadushan: We may have to cast it then.

ThisaruGuruge: Yes, we can use casting or typeguard, whatever suits here. Using `toString()` is not an ideal way to handle these cases. There are many complexities in using `toString()`.

ballerina/common_utils.bal
```diff
@@ -166,6 +166,15 @@ isolated function getOfType(__Type schemaType) returns __Type {
...
 }
 
+isolated function getOfTypeName(__Type schemaType) returns string {
+    __Type? ofType = schemaType?.ofType;
+    if ofType is () {
+        return <string>schemaType.name;
+    } else {
+        return <string>getOfType(ofType).name;
+    }
+}
```
ThisaruGuruge, line 176: Can't this be simplified like this?

```suggestion
isolated function getOfTypeName(__Type schemaType) returns string {
    return <string>getOfType(schemaType).name
}
```

We are using `getOfType()` function but here we are trying to do the first level of unwrapping inside this function, which is redundant.

ThisaruGuruge (approved):

PR #657, opened 2022-03-18
Improve GraphQL Listener for 'Subscriptions' Support 

DimuthuMadushan:
## Purpose

Fixes: [#2780](https://github.com/ballerina-platform/ballerina-standard-library/issues/2780)

## Examples
N/A

## Checklist
- [x] Linked to an issue
- [ ] ~Updated the changelog~
- [ ] Added tests
- [ ] ~Updated the spec~


ballerina/listener.bal
```diff
@@ -84,6 +114,12 @@ public class Listener {
         if result is error {
             return error Error("Error occurred while starting the service", result);
         }
+        if self.wsListener is websocket:Listener {
+            result = (<websocket:Listener>self.wsListener).'start();
```
ThisaruGuruge, line 118: @MaryamZi What should be the better way to stop another listener like this?

* To use type-guard and cast
* To define a local variable and then use that

ballerina/websocket_service.bal
```diff
@@ -0,0 +1,92 @@
...
+        lock {
+            parser:OperationNode|ErrorDetail node = validateSubscriptionPayload(data, self.engine);
+            if node is parser:OperationNode {
+                self.clientsMap[caller.getConnectionId()] = [caller, node];
+
+                stream<any, io:Error?> sourceStream = new(); // receiving stream from "getSubscriptionResponse(...)" -> Not implemented yet
+                any|io:Error? next = sourceStream.iterator().next();
+
+                while next !is io:Error? {
+                    check broadcast(self.engine, self.schema, self.context, next, self.clientsMap);
```
ThisaruGuruge, line 53: What happens when this returns an error? I think we should handle it internally and send a proper GraphQL error response.

ballerina/listener.bal
MaryamZi: The reason why the type is not narrowed for `self.wsListener` is that the value (and therefore the type) may change to be something else from when the check was done vs when the value is actually used in a context where it is expected to be of the tested type. So here for example, we can't guarantee that `self.wsListener` is a `websocket:Listener` in L118. Due to the same reason, using a cast is not generally safe since the `self.wsListener` value may have changed.

Of course, the author may know that this value will never change and may consider it safe. But assigning to a local variable and then doing the type test is guaranteed to be safe.

Moreover, we do the field access just once if we use a local variable, and we can get rid of an unnecessary access.

ballerina/websocket_service.bal
DimuthuMadushan: `broadcast()` is returning a `websocket:Error` like in line 57. This error is the return type of `writeTextMessage()` function in websocket. Do we need to handle it internally?

ballerina/listener.bal
```diff
@@ -53,13 +56,30 @@ public class Listener {
         Engine engine = check new (schema, maxQueryDepth);
         attachServiceToEngine(s, engine);
 
+        if schema.hasKey(SUBSCRIPTION_FIELD) {
+            websocket:Listener|error wsListener = new(self.httpListener);
+            if wsListener is error {
+                return error Error("Listener initialization failed", wsListener);
```
ThisaruGuruge, line 62: Let's make it clear that Websocket listener creation is the one that failed.

ballerina/listener.bal
```diff
@@ -74,6 +94,16 @@ public class Listener {
                 return error Error("Error occurred while detaching the service", result);
             }
         }
+        websocket:Listener? wsListener = self.wsListener;
+        if wsListener is websocket:Listener {
+            UpgradeService? wsService = getWebsocketServiceFromGraphqlService(s);
```
ThisaruGuruge, line 99: Why do we need an interop function to get the websocket service? It is a class variable, right?

ballerina/websocket_service.bal
ThisaruGuruge: Let's keep this as it is for now then. But we should look at how other libraries handle these kinds of scenarios. I think we should send a GraphQL specific error here.

ballerina/listener.bal
DimuthuMadushan: This is to get the `UpgradeService`. It's not a class variable.

ballerina/listener_utils.bal
```diff
@@ -355,3 +364,12 @@ isolated function getHttpServiceFromGraphqlService(Service s) returns HttpServic
 @java:Method {
     'class: "io.ballerina.stdlib.graphql.runtime.engine.ListenerUtils"
 } external;
+
+isolated function attachUpgradeServiceToGraphqlService(Service s, UpgradeService wsService) = @java:Method {
```
ThisaruGuruge, line 368: Shouldn't this be `attachWebsocketServiceToGraphqlService` (If we follow that same pattern)?

ThisaruGuruge (approved):

PR #676, opened 2022-04-05
Integrate GraphiQL Client into GraphQL Package

DimuthuMadushan:
## Purpose

Fixes: [#1936](https://github.com/ballerina-platform/ballerina-standard-library/issues/1936)

## Examples
GraphiQL client can be enable to a GraphQL service  as follow by providing `graphiql` configurations.
```ballerina
graphql:ServiceConfig {
    graphiql: {
        enable: true,
        path: "/ballerina/graphiql"
    }
}
service on new graphql:Listener(4000) {
    resource function get greeting() returns string {
        return "Hello, World!";
    }
}
```
GraphiQL client will be served for this service at `http://localhost:4000/ballerina/graphiql`.

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [x] Updated the spec


native/src/main/java/io/ballerina/stdlib/graphql/runtime/engine/ListenerUtils.java
```diff
@@ -18,25 +18,114 @@
...
+    public static Object validateGraphiQLPath(BString path) {
+        String uri = SAMPLE_URL + path;
+        try {
+            new URL(uri).toURI();
+            return null;
+        } catch (URISyntaxException e) {
+            return createError("Invalid path provided for GraphiQL client", ERROR_TYPE);
+        } catch (MalformedURLException e) {
+            return createError("Invalid path provided for GraphiQL client", ERROR_TYPE);
+        }
```
ThisaruGuruge, line 86: Since we are returning the same error, we don't need two separate catch blocks, right?

I'd also suggest to add the message from the exception for a clearer error.

```suggestion
        } catch (URISyntaxException | MalformedURLException e) {
            return createError("Invalid path provided for GraphiQL client", ERROR_TYPE);
        }
```

ThisaruGuruge (commented):
Let's add the proposal with this as well.

ballerina/types.bal
```diff
@@ -52,6 +52,15 @@ public type CorsConfig record {|
     *http:CorsConfig;
 |};
 
+# Represent GraphiQL client configurations
+#
+# + enable - Status of the client
+# + path - Path for the client
+public type GraphiQL record {|
+    boolean enable;
```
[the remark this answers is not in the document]
DimuthuMadushan: Since the `graphiql` is an optional field in Service Configuration record, do we need to add this default value here?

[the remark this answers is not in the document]
shafreenAnfar: Adding it improves the readability right ? By looking at the config one can explicitly know it is not enabled.

ballerina/listener.bal
```diff
@@ -60,6 +61,17 @@ public class Listener {
...
         }
+
+        if graphiql is GraphiQL && graphiql.enable {
+            check validateGraphiQLPath(graphiql.path);
+            string gqlServiceBasePath = name is () ? "" : getBasePath(name);
+            HttpService graphiqlService = getGraphiQLService(serviceConfig, gqlServiceBasePath);
+            attachGraphiQLServiceToGraphqlService(s, graphiqlService);
+            result = self.httpListener.attach(graphiqlService, graphiql.path);
+            if result is error {
+                return error Error("Error occured while attaching the GraphiQL service", result);
```
shafreenAnfar, line 72: "Error occured while attaching the GraphiQL service" -> "Error occured while attaching the GraphiQL endpoint" ?

shafreenAnfar: occured -> occurred

docs/spec/spec.md
```diff
@@ -3,7 +3,7 @@
 _Owners_: @shafreenAnfar @DimuthuMadushan @ThisaruGuruge  
 _Reviewers_: @shafreenAnfar @DimuthuMadushan @ldclakmal  
 _Created_: 2022/01/06  
-_Updated_: 2022/02/09  
+_Updated_: 2022/04/05
```
shafreenAnfar, line 6:
```suggestion
_Updated_: 2022/04/05  
```

ballerina/types.bal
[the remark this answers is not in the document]
ThisaruGuruge: > Since the `graphiql` is an optional field in Service Configuration record, do we need to add this default value here?

Adding to what @shafreenAnfar mentioned, I think since we have default values for each field, it does not need to be an optional field either. We can make it a required field but disabled it by default.

ballerina/listener.bal
```diff
@@ -49,17 +49,28 @@ public class Listener {
         __Schema schema = check createSchema(s);
         addDefaultDirectives(schema);
         GraphqlServiceConfig? serviceConfig = getServiceConfig(s);
+        Graphiql graphiql = getGraphiQLConfig(serviceConfig);
```
ThisaruGuruge, line 52: Shall we check where we missed to fix this?

```suggestion
        Graphiql graphiql = getGraphiqlConfig(serviceConfig);
```

ThisaruGuruge (approved):

ThisaruGuruge (approved):

PR #683, opened 2022-04-11
Improve Schema Generation for Subscriptions 

Nuvindu:
## Purpose

Fixes:

## Checklist
- [ ] Linked to an issue
- [ ] Updated the changelog
- [ ] Added tests
- [ ] Updated the spec


ballerina/tests/35_subscriptions.bal
```diff
@@ -0,0 +1,169 @@
+import ballerina/test;
+import ballerina/websocket;
+import ballerina/lang.value;
```
DimuthuMadushan, line 3: shall we add an empty line here and after each test case.

ballerina/tests/35_subscriptions.bal
```diff
@@ -0,0 +1,169 @@
...
+    check wsClient->writeTextMessage(document);
+    string textResponse = check wsClient->readTextMessage();
+    int i = 1;
+    while (i < 4) {
+        json actualPayload = check value:fromJsonString(textResponse);
+        json expectedPayload = { data : { messages: i }};
+        i += 1;
+        assertJsonValuesWithOrder(actualPayload, expectedPayload);
+        textResponse = check wsClient->readTextMessage();
+    }
```
DimuthuMadushan, line 20:
```suggestion
    foreach int i in 1..< 4 {
        json actualPayload = check value:fromJsonString(textResponse);
        json expectedPayload = { data : { messages: i }};
        assertJsonValuesWithOrder(actualPayload, expectedPayload);
        textResponse = check wsClient->readTextMessage();
    }
```

ballerina/tests/35_subscriptions.bal
```diff
@@ -0,0 +1,169 @@
...
+
+        json actualPayload2 = check value:fromJsonString(textResponse2);
+        json expectedPayload2 = { data : { stringMessages: i.toString() }};
+
+        i += 1;
+        assertJsonValuesWithOrder(actualPayload2, expectedPayload2);
+        textResponse2 = check wsClient2->readTextMessage();
+    }
+
+
```
DimuthuMadushan, line 75: extra empty lines?

ballerina/tests/35_subscriptions.bal
```diff
@@ -0,0 +1,169 @@
...
+    fragment peopleFields on Subscription {
+        ...personFields 
+    }
+    fragment personFields on Subscription {
+        people {
+            name
+            age       
+        } 
+    }
+    `;
```
DimuthuMadushan, line 156: Add to a separate file.

ballerina/tests/test_services.bal
```diff
@@ -1142,3 +1142,38 @@ service /nullable_inputs on basicListener {
         return account.number;
     }
 }
+
+service /subscriptions on basicListener {
+
+    isolated resource function get name() returns string {
+        return "Walter White";
+    }
```
DimuthuMadushan, line 1150: add an empty line after each resource function.

ballerina/websocket_service.bal
```diff
@@ -14,9 +14,8 @@
 // specific language governing permissions and limitations
 // under the License.
 
-import ballerina/io;
 import ballerina/websocket;
-
+// import ballerina/io;
```
DimuthuMadushan, line 18: Remove unused imports

ballerina/websocket_service.bal
```diff
@@ -25,43 +24,64 @@ isolated service class WsService {
...
         self.context = new();
-        self.clientsMap = {};
     }
                 
     isolated remote function onClose(websocket:Caller caller, int statusCode, string reason) {
         lock {
-            _ = self.clientsMap.remove(caller.getConnectionId());
+            // _ = self.clientsMap.remove(caller.getConnectionId());
         }
     }
```
DimuthuMadushan, line 38: Remove this if there are no tasks to do with WebSocket `onClose()` function.

ballerina/websocket_service.bal
```diff
@@ -25,43 +24,64 @@ isolated service class WsService {
...
 }
 
 isolated function validateSubscriptionPayload(string text, Engine engine) returns parser:OperationNode|ErrorDetail {
     json payload = text.toJson();
     var document = payload.query;
+    if document is error {
+        document = payload;
+    }
```
DimuthuMadushan, line 84: Do we need to assign payload to the `document` here?  The error is already handled.

ballerina/websocket_service.bal
```diff
@@ -78,15 +98,7 @@ isolated function validateSubscriptionPayload(string text, Engine engine) return
...
-        ExecutorVisitor executor = new(engine, schema, context, {}, result);
-        websocket:Caller caller = callerDetails[0];
-        parser:OperationNode operationNode = callerDetails[1];
-        OutputObject outputObject = executor.getExecutorResult(operationNode);
-        if outputObject.hasKey(DATA_FIELD) || outputObject.hasKey(ERRORS_FIELD) {
-            check caller->writeTextMessage(outputObject.toString());
-        }
-    }
-}
+isolated function getSubscriptionResponse(Engine engine, __Schema schema, Context context, string operationName) returns stream<any,error?>|error {
```
DimuthuMadushan, line 101: Isn't this exceeding the maximum character length? The maximum number of characters per line should be under 120.

ballerina/websocket_service.bal
```diff
@@ -78,15 +98,7 @@ isolated function validateSubscriptionPayload(string text, Engine engine) return
...
+    ExecutorVisitor executor = new(engine, schema, context, {}, null);
+    return <stream<any,error?>|error> getSubscriptionResult(executor, operationName);
+}
```
DimuthuMadushan, line 104: Add an empty line here.

native/src/main/java/io/ballerina/stdlib/graphql/runtime/engine/Engine.java
```diff
@@ -171,11 +174,57 @@ private static void getExecutionResult(ExecutionContext executionContext, BObjec
...
+        BObject engine = visitor.getObjectValue(ENGINE_FIELD);
+        BObject service = (BObject) engine.getNativeData(GRAPHQL_SERVICE_OBJECT);
+        ServiceType serviceType = (ServiceType) service.getType();
+        UnionType typeUnion = TypeCreator.createUnionType(PredefinedTypes.TYPE_STREAM, PredefinedTypes.TYPE_ERROR);
+        for (ResourceMethodType resourceMethod : serviceType.getResourceMethods()) {
+            if (Objects.equals(String.valueOf(operationName), resourceMethod.getResourcePath()[0])) {
+                if (service.getType().isIsolated() && service.getType().isIsolated(resourceMethod.getName())) {
+                    env.getRuntime()
+                            .invokeMethodAsyncConcurrently(service, resourceMethod.getName(), null,
+                                    null, new Callback() {
```
DimuthuMadushan, line 195: Implement the `callback` separately.

native/src/main/java/io/ballerina/stdlib/graphql/runtime/engine/Engine.java
```diff
@@ -171,11 +174,57 @@ private static void getExecutionResult(ExecutionContext executionContext, BObjec
...
+                                        @Override
+                                        public void notifyFailure(BError bError) {
+
+                                            subscriptionFutureResult.complete(bError);
+                                        }
+                                    }, null, typeUnion, (Object) null);
+                }
+            }
+        }
+
```
DimuthuMadushan, line 227: Extra empty line?

native/src/main/java/io/ballerina/stdlib/graphql/runtime/schema/FieldFinder.java
```diff
@@ -232,8 +258,10 @@ private void getFieldsFromInputObjectType(SchemaType schemaType, Field field) {
     private void getFieldsFromServiceType(SchemaType schemaType) {
         ServiceType serviceType = (ServiceType) schemaType.getBalType();
         for (ResourceMethodType resourceMethod : serviceType.getResourceMethods()) {
-            String[] resourcePath = resourceMethod.getResourcePath();
-            schemaType.addField(getFieldsFromResourceMethodType(resourceMethod, resourcePath));
+            if (Objects.equals(resourceMethod.getAccessor(), GET_ACCESSOR)) {
```
ThisaruGuruge, line 261: This check is redundant as this method is used to get the fields of a service-type object. They all have the `get` accessor in the resource methods. (Only the top-level GraphQL service can have the `subscribe` accessor).

native/src/main/java/io/ballerina/stdlib/graphql/runtime/schema/TypeFinder.java
```diff
@@ -116,6 +125,13 @@ private void getTypesFromRemoteMethod(RemoteMethodType remoteMethod) {
         getSchemaTypeFromBalType(remoteMethod.getType().getReturnType());
     }
 
+    private void getTypesFromSubscriptionResourceMethod(ResourceMethodType resourceMethod) {
+        getInputTypesFromMethod(resourceMethod);
+        Type type = resourceMethod.getType().getReturnType();
+        StreamType streamType = (StreamType) type;
```
ThisaruGuruge, line 131:
```suggestion
        StreamType streamType = (StreamType) resourceMethod.getType().getReturnType();
```

native/src/main/java/io/ballerina/stdlib/graphql/runtime/schema/FieldFinder.java
```diff
@@ -269,6 +306,15 @@ private void setTypeForField(MethodType method, SchemaField schemaField) {
         }
     }
 
+    private void setTypeForField(Type resourceReturnType, SchemaField schemaField) {
```
ThisaruGuruge, line 309: Why can't we use `setTypeForField(MethodType method, SchemaField schemaField)` here? We are basically doing the same thing, right?

ballerina/websocket_service.bal
```diff
@@ -14,54 +14,65 @@
...
-                self.clientsMap[caller.getConnectionId()] = [caller, node];
-
-                stream<any, io:Error?> sourceStream = new(); // receiving stream from "getSubscriptionResponse(...)" -> Not implemented yet
-                any|io:Error? next = sourceStream.iterator().next();
-
-                while next !is io:Error? {
-                    check broadcast(self.engine, self.schema, self.context, next, self.clientsMap);
-                    next = sourceStream.iterator().next();
-                }
+                parser:Selection selection = node.getSelections()[0];
```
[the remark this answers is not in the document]
Nuvindu: I wrote it to get the field node from the operation node. Since it is a subscription query, there can be only one element in the selection nodes array. Although it can be a fragment node, a while loop is iterated until it reaches the top level field node.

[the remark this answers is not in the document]
ThisaruGuruge: Then the Parser should validate the subscription operation to have exactly one field.

ballerina/websocket_service.bal
```diff
@@ -14,54 +14,65 @@
...
+                            check caller->writeTextMessage(coercedOutputObject.toString());
+                        }
+                        next = sourceStream.iterator().next();
+                    }
+                } else {
+                    check caller->writeTextMessage(sourceStream.toString());
+                }                
+            } else {
+                check caller->writeTextMessage((<ErrorDetail>node).message);
             }
```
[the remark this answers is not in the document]
Nuvindu: Then it will call for correct graphql requests also

ballerina/websocket_service.bal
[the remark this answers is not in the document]
DimuthuMadushan: It seems `fragment node` validation and the `duplicate field removing` should do before the `subscription operation` validation. In that case, it's better to validate using a visitor. @ThisaruGuruge WDYT?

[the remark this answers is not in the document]
ThisaruGuruge: +1

ballerina/modules/parser/parser.bal
```diff
@@ -154,13 +154,22 @@ public class Parser {
...
                 FieldNode fieldNode = check self.addSelectionToNode(parentNode);
-                parentNode.addSelection(fieldNode);
+                parentNode.addSelection(fieldNode);                
+                if parentNode is OperationNode {
+                    if parentNode.getKind() == OPERATION_SUBSCRIPTION {
+                        hasRootField = true;
+                    }
+                }
+            } else {
+                string message = "Subscription operations must have exactly one root field.";
```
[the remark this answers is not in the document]
Nuvindu: Okay. I will modify with the exact error message.

ballerina/websocket_service.bal
```diff
@@ -14,60 +14,70 @@
...
+                stream<any,error?>|error sourceStream = getSubscriptionResponse(self.engine, self.schema, 
+                                                                                self.context, fieldNode);                 
+                if sourceStream is stream<any,error?> {
+                    record{|any value;|}|error? next = sourceStream.iterator().next();
+                    while next !is error? {
+                        ExecutorVisitor executor = new(self.engine, self.schema, self.context, {}, next.value);
+                        OutputObject outputObject = executor.getExecutorResult(node);
+                        ResponseFormatter responseFormatter = new(self.schema);
+                        OutputObject coercedOutputObject = responseFormatter.getCoercedOutputObject(outputObject, 
+                                                                                                    node);
```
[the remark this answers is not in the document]
Nuvindu: Then the line 51 will have 121 characters.

ballerina/engine.bal
```diff
@@ -98,6 +98,12 @@ isolated class Engine {
             return getOutputObjectFromErrorDetail(errors);
         }
 
+        SubscriptionVisitor subscriptionVisitor = new(document);
+        ErrorDetail[]? subscriptionErrors = subscriptionVisitor.validate();
```
DimuthuMadushan, line 102: Can't we use the same `errors` variable instead of `subscriptionErrors`?

ballerina/root_field_visitor.bal
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }
+
+    public isolated function visitField(parser:FieldNode fieldNode, anydata data = ()) {
+        // Do nothing
+    }
+
+    public isolated function visitFragment(parser:FragmentNode fragmentNode, anydata data = ()) {
+        // Do nothing
+    }
```
DimuthuMadushan, line 60:
```suggestion
    public isolated function visitSelection(parser:Selection selection, anydata data = ()) {
        if selection is parser:FragmentNode {
            self.visitFragment(selection);
        } else {
            self.visitField(selection);
        }
    }

    public isolated function visitField(parser:FieldNode fieldNode, anydata data = ()) {
        self.fieldNodes.push(fieldNode);
    }

    public isolated function visitFragment(parser:FragmentNode fragmentNode, anydata data = ()) {
        self.visitSelection(fragmentNode.getSelections()[0]);
    }
```

ballerina/subscription_visitor.bal
```diff
@@ -0,0 +1,103 @@
...
+    public isolated function visitSelection(parser:Selection selection, anydata data = ()) {
+        if selection is parser:FragmentNode {
+            self.visitFragment(<parser:FragmentNode>selection, data);
+        } else {
+            self.visitField(<parser:FieldNode>selection, data);
+        }
+    }
+
+    public isolated function visitField(parser:FieldNode fieldNode, anydata data = ()) {
+        if fieldNode.getName().substring(0,2) == "__" {
```
DimuthuMadushan, line 66:
```suggestion
        if fieldNode.getName() == SCHEMA_FIELD {
```
can't we use this here like in `ExecutorVisitor`?

ballerina/subscription_visitor.bal
```diff
@@ -0,0 +1,103 @@
...
+    public isolated function visitArgument(parser:ArgumentNode argumentNode, anydata data = ()) {
+        // Do nothing
+    }
+
+    public isolated function addErrorDetail(parser:Selection selection, string operationName) {
+        string message = operationName != "<anonymous>"  
+                         ? string`Subscription "${operationName}" must select only one top level field.`
+                         : string`Anonymous Subscription must select only one top level field.`;   
+        if selection is parser:FragmentNode {
+            ErrorDetail errorDetail = getErrorDetailRecord(message, (<parser:FragmentNode>selection).getLocation());
```
DimuthuMadushan, line 88: Do we need to cast the `selection`?

ballerina/tests/resources/documents/subscriptions_with_introspections.graphql
```diff
@@ -0,0 +1,3 @@
+subscription sub { 
+    __typename 
+}
```
DimuthuMadushan, line 3: Shall we add a test case for the following scenario as well?
`{ messages { __typename { name } } }`

Nuvindu: Sure, I'll add that as well

ballerina/engine.bal
```diff
@@ -98,6 +98,12 @@ isolated class Engine {
             return getOutputObjectFromErrorDetail(errors);
         }
 
+        SubscriptionVisitor subscriptionVisitor = new(document);
```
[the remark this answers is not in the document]
DimuthuMadushan: Yes. This should be done after the `DuplicateFieldRemover`.

ballerina/tests/35_subscriptions.bal
```diff
@@ -150,6 +138,30 @@ isolated function testSubscriptionWithVariables() returns error? {
...
+}
+isolated function testInvalidSubscription() returns error? {
+    string document = string`subscription { invalidField }`;
+    string url = "ws://localhost:9091/subscriptions";
+    websocket:Client wsClient = check new(url);
+    check writeWebSocketTextMessage(document, wsClient);
+    json expectedPayload = check getJsonContentFromFile("subscription_invalid_field.json");
+    check validateWebSocketResponse(wsClient, expectedPayload);
+}
+
```
DimuthuMadushan, line 164: Do we have tests for invalid WebSocket payload validation?

ballerina/tests/35_subscriptions.bal
```diff
@@ -0,0 +1,199 @@
...
+    foreach int i in 1..< 3 {
+        json expectedPayload = { data : { filterValues: i }};
+        check validateWebSocketResponse(wsClient, expectedPayload);
+    }
+}
+
+@test:Config {
+    groups: ["introspection", "typename", "subscriptions"]
+}
+isolated function testSubscriptionWithIntrsopectionInFields() returns error? {
```
DimuthuMadushan, line 144:
```suggestion
isolated function testSubscriptionWithIntrospectionInFields() returns error? {
```

ballerina/root_field_visitor.bal
```diff
@@ -0,0 +1,63 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import graphql.parser;
+
+class RootFieldVisitor {
+    *parser:Visitor;
+
+    private parser:FieldNode[] fieldNodes;
```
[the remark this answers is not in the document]
Nuvindu: If the `fieldNode` variable is nullable, an error handling is required to assure it is not not null before it has been used in another function. Since the `operationNode` is already validated there is no need of error handling required here. I used an array here, because it can be used to return non nullable `fieldNode`.

ballerina/tests/35_subscriptions.bal
Nuvindu: Currently, no. I'll add them.

ballerina/subscription_visitor.bal
```diff
@@ -0,0 +1,104 @@
...
+        foreach parser:OperationNode operationNode in operations {
+            self.visitOperation(operationNode);
+        }
+    }
+
+    public isolated function visitOperation(parser:OperationNode operationNode, anydata data = ()) {
+        if operationNode.getKind() == parser:OPERATION_SUBSCRIPTION {
+            parser:Selection[] selections = operationNode.getSelections();
+            if selections.length() > 1 {
+                self.addErrorDetail(selections[1], operationNode.getName());
```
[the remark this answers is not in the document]
Nuvindu: In the reference implementation, if a subscription operation has multiple fields, the error location will be the beginning of the invalid field/s.

ballerina/root_field_visitor.bal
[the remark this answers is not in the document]
ThisaruGuruge: No, that's not the way to handle it.

A nullable variable does not need error handling. It only needs to check whether the value is null or not. In other words, if we have a value or not. Using an array to bypass that check does not solve the original issue. What if the array is an empty one? The same issue occurs.

Even from the memory point of view, this has a negative impact.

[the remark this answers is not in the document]
DimuthuMadushan: BTW can `fieldNode` be nullable at this point? Doesn't it validate at the `subscription_visitor`?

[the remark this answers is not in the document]
ThisaruGuruge: It should not be null. But the point is, that we cannot initialize it to the exact type in the beginning. So we have to use a nullable type. Otherwise, an option is not to define a class variable but use a local variable. But I don't think that would work in this case.

ballerina/tests/utils.bal
```diff
@@ -112,8 +112,12 @@ isolated function writeWebSocketTextMessage(string document, websocket:Client ws
 }
 
 isolated function validateWebSocketResponse(websocket:Client wsClient,
-                                            json expectedPayload) returns websocket:Error?|error {
+                                            json|string expectedPayload) returns websocket:Error?|error {
```
DimuthuMadushan, line 115: Do we need to add `string` type here?

ballerina/subscription_visitor.bal
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+
+    public isolated function visitSelection(parser:Selection selection, anydata data = ()) {
+        if selection is parser:FragmentNode {
+            self.visitFragment(<parser:FragmentNode>selection, data);
+        } else {
+            self.visitField(<parser:FieldNode>selection, data);
+        }
+    }
```
DimuthuMadushan, line 63: Shall we remove the casting here?

ballerina/subscription_visitor.bal
```diff
@@ -88,8 +88,8 @@ class SubscriptionVisitor {
         if selection is parser:FragmentNode {
             ErrorDetail errorDetail = getErrorDetailRecord(message, selection.getLocation());
             self.errors.push(errorDetail);
-        } else {
-            ErrorDetail errorDetail = getErrorDetailRecord(message, (<parser:FieldNode>selection).getLocation());
+        } else if selection is parser:FieldNode {
+            ErrorDetail errorDetail = getErrorDetailRecord(message, selection.getLocation());
             self.errors.push(errorDetail);
         }
```
DimuthuMadushan, line 94: Missing `else` block?

ballerina/tests/utils.bal
Nuvindu: It was added because I can check both invalid error messages (which are in strings) and valid json responses using the same method

ballerina/subscription_visitor.bal
Nuvindu: If the `else`  block is used instead of `else if` , we have to use casting in the next line, to get the location from the selection node. Since the `Selection` node can only be either `FragmentNode` or `FieldNode` , it is not necessary to use `else` block here.

ballerina/tests/utils.bal
Nuvindu: No, I'll remove it.

DimuthuMadushan: [`json`](https://ballerina.io/learn/by-example/json-type.html) type can accept `string` value also.

ballerina/subscription_visitor.bal
DimuthuMadushan: Here we need to use the `else` block after that lang change. right? @ThisaruGuruge

ballerina/subscription_visitor.bal
```diff
@@ -94,11 +94,11 @@ class SubscriptionVisitor {
...
     }
 
-    public isolated function addIntrospectionErrorDetail(parser:Selection selection, string operationName) {
+    public isolated function addIntrospectionErrorDetail(parser:FieldNode selection, string operationName) {
```
DimuthuMadushan, line 97: Change the parameter name as well.

ballerina/subscription_visitor.bal
ThisaruGuruge: Since it does not return a value, I think we don't need it. @Nuvindu Let's check whether it complains of any compilation errors here.

Nuvindu: @ThisaruGuruge I checked with the latest version. No any compilation error occurred here.

DimuthuMadushan (approved):
LGTM!

ballerina/websocket_service.bal
```diff
@@ -14,56 +14,60 @@
 // specific language governing permissions and limitations
 // under the License.
 
-import ballerina/io;
 import ballerina/websocket;
-
 import graphql.parser;
+import ballerina/lang.value;
```
ThisaruGuruge, line 19: Shall we re-order the imports here?

native/src/main/java/io/ballerina/stdlib/graphql/runtime/engine/Engine.java
```diff
@@ -178,4 +183,33 @@ private static void getExecutionResult(ExecutionContext executionContext, BObjec
...
+        CallbackHandler callbackHandler = new CallbackHandler(subscriptionFutureResult);
+        ExecutionContext executionContext = new ExecutionContext(env, visitor, callbackHandler, SUBSCRIPTION);
+        BString fieldName = node.getStringValue(NAME_FIELD);
+        BObject engine = visitor.getObjectValue(ENGINE_FIELD);
+        BObject service = (BObject) engine.getNativeData(GRAPHQL_SERVICE_OBJECT);
+        ServiceType serviceType = (ServiceType) service.getType();
+        UnionType typeUnion = TypeCreator.createUnionType(PredefinedTypes.TYPE_STREAM, PredefinedTypes.TYPE_ERROR);
+        for (ResourceMethodType resourceMethod : serviceType.getResourceMethods()) {
+            if (Objects.equals(String.valueOf(fieldName), resourceMethod.getResourcePath()[0]) &&
+                Objects.equals(resourceMethod.getAccessor(), SUBSCRIBE_ACCESSOR)) {
```
ThisaruGuruge, line 199: Let's not use `Objects.equals()` method, to maintain consistency. Please check other places as well.

```suggestion
            if (SUBSCRIBE_ACCESSOR.equals(resourceMethod.getAccessor()) &&
                fieldName.getValue().equals(resourceMethod.getResourcePath()[0]) {
```

native/src/main/java/io/ballerina/stdlib/graphql/runtime/engine/SubscriptionCallback.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package io.ballerina.stdlib.graphql.runtime.engine;
+
+import io.ballerina.runtime.api.Future;
+import io.ballerina.runtime.api.async.Callback;
+import io.ballerina.runtime.api.values.BError;
+
+/**
+ * Callback class for the async invocation of the Ballerina subscriptions.
```
ThisaruGuruge, line 26:
```suggestion
 * Callback class for the async invocation of the Ballerina GraphQL subscription resources.
```

native/src/main/java/io/ballerina/stdlib/graphql/runtime/schema/FieldFinder.java
```diff
@@ -94,6 +103,21 @@ public SchemaType populateFieldsForMutationType() {
         return mutationType;
     }
 
+    public SchemaType populateFieldsForSubscriptionType() {
+        SchemaType subscriptionType = null;
+        if (this.typeMap.containsKey(SUBSCRIPTION)) {
+            subscriptionType = this.typeMap.remove(SUBSCRIPTION);
+            ServiceType serviceType = (ServiceType) subscriptionType.getBalType();
+            for (ResourceMethodType resourceMethodType : serviceType.getResourceMethods()) {
+                if (Objects.equals(resourceMethodType.getAccessor(), SUBSCRIBE_ACCESSOR)) {
```
ThisaruGuruge, line 112: Ditto

```suggestion
                if (SUBSCRIBE_ACCESSOR.equals(resourceMethodType.getAccessor())) {
```

native/src/main/java/io/ballerina/stdlib/graphql/runtime/schema/TypeFinder.java
```diff
@@ -85,7 +89,12 @@ private void getDefaultTypes() {
     private void createTypesFromRootService() {
         this.createSchemaType(QUERY, TypeKind.OBJECT, serviceType);
         for (ResourceMethodType resourceMethod : this.serviceType.getResourceMethods()) {
-            getTypesFromResourceMethod(resourceMethod, resourceMethod.getResourcePath());
+            if (Objects.equals(resourceMethod.getAccessor(), GET_ACCESSOR)) {
```
ThisaruGuruge, line 92: Same as above
```suggestion
            if (GET_ACCESSOR.equals(resourceMethod.getAccessor()) {
```

ballerina/websocket_service.bal
```diff
@@ -25,68 +24,66 @@ isolated service class WsService {
...
-                return validationResult;
-            }
-            return {message: validationResult.toJsonString()};
-        }
+    json|error payload = value:fromJsonString(text);
+    if payload is error {
+        return {message: "Invalid Subscription Payload"};
+    }
+    json|error document = payload.query;
+    if !(document is string) || document == "" {
```
DimuthuMadushan, line 70:
```suggestion
    if document !is string || document == "" {
```

ballerina/websocket_service.bal
```diff
@@ -25,68 +24,66 @@ isolated service class WsService {
...
+        return {message: "Query not found"};
+    }
+    json|error variables = payload.variables;
+    if !(variables is map<json>) && variables != () {
```
DimuthuMadushan, line 74:
```suggestion
    if variables !is map<json> && variables != () {
```

native/src/main/java/io/ballerina/stdlib/graphql/runtime/schema/TypeFinder.java
```diff
@@ -116,6 +126,26 @@ private void getTypesFromRemoteMethod(RemoteMethodType remoteMethod) {
...
+        } else if (resourceMethod.getType().getReturnType().getTag() == UNION_TAG) {
+            List<Type> memberTypes = ((UnionType) resourceMethod.getType().getReturnType()).getMemberTypes();
+            for (Type memberType: memberTypes) {
+                if (memberType.getTag() == STREAM_TAG) {
+                    streamType = (StreamType) memberType;
+                }
+            }
+        }
+        if (streamType == null) {
+            throw createError("Invalid return type for subscription: Unable to find the stream", ERROR_TYPE);
```
ThisaruGuruge, line 143:
```suggestion
            throw createError("Invalid return type for subscription: return type must consist a stream", ERROR_TYPE);
```

ballerina/websocket_service.bal
```diff
@@ -25,68 +24,73 @@ isolated service class WsService {
...
+        json errorMessage = {errors: [{message: "Invalid subscription payload"}]};
+        return {message: errorMessage.toJsonString()};
+    }
+    json|error document = payload.query;
+    if document !is string || document == "" {
+        json errorMessage = {errors: [{message: "Query not found"}]};
+        return {message: errorMessage.toJsonString()};
     }
-    return {message: "Query not found"};
+    json|error variables = payload.variables;
```
DimuthuMadushan, line 75: Since the `variables field` is optional in standard GraphQL requests, this can be an `error` even for the valid requests.

ballerina/tests/test_services.bal
```diff
@@ -1142,3 +1229,60 @@ service /nullable_inputs on basicListener {
...
+        return names.toStream();
+    }
+
+    isolated resource function subscribe messages() returns stream<int>|error {
+        int[] intArray = [1, 2, 3, 4, 5];
+        return intArray.toStream();
+    }
+
+    isolated resource function subscribe stringMessages() returns stream<string?, error?> {
+        string[] stringArray = ["1", "2", "3", "4", "5"];
```
DimuthuMadushan, line 1249: Since the resource returns a `string?` stream, shall we add a `null` value also?

ballerina/tests/test_services.bal
```diff
@@ -1142,3 +1142,61 @@ service /nullable_inputs on basicListener {
         return account.number;
     }
 }
+
+public string[] namesArray = ["Walter", "Skyler"];
```
DimuthuMadushan, line 1146: Shall we move this to inside the service?

ballerina/tests/test_services.bal
```diff
@@ -1142,3 +1142,61 @@ service /nullable_inputs on basicListener {
...
+            }
+        }
+        return filteredArray.toStream();
+    }
+
+    isolated resource function subscribe values() returns stream<int> {
+        int[] array = [];
+        foreach int i in 1 ..< 2 {
+            int _ = array.remove(0);
+        }
```
DimuthuMadushan, line 1195: What's the purpose of using `foreach` loop?

ballerina/tests/test_services.bal
Nuvindu: It was there to make a resource function non-isolated. Then the `invokeMethodAsyncSequentially()` method can be tested. If it was inside the service, I think the whole service is going to be non-isolated. Then the `invokeMethodAsyncConcurrently()` method will not run. To run both methods, array was kept there

ballerina/tests/test_services.bal
Nuvindu: To return an unexpected error here. So the `notifyFailure()` method in the `SubscriptionCallback` class will be invoked.

Nuvindu: I will remove it. It is not necessary

ballerina/tests/test_services.bal
```diff
@@ -1185,4 +1187,14 @@ service /subscriptions on basicListener {
         }
         return filteredArray.toStream();
     }
+
+    isolated resource function subscribe values() returns stream<int> {
+        int[] array = [];
+        int _ = array.remove(0);
```
DimuthuMadushan, line 1193: Any reason to include this line? In which test case, this resource function is used?

ballerina/websocket_service.bal
```diff
@@ -25,68 +24,73 @@ isolated service class WsService {
...
-    }
-                
-    isolated remote function onClose(websocket:Caller caller, int statusCode, string reason) {
-        lock {
-            _ = self.clientsMap.remove(caller.getConnectionId());
-        }
     }
 
     isolated remote function onTextMessage(websocket:Caller caller, string data) returns websocket:Error? {
         lock {
```
DimuthuMadushan, line 35: I think we can remove this now. This was added due to the `client map` that we have in the earlier design. Shall we try this without `lock`?

ballerina/tests/test_services.bal
Nuvindu: It was there to return an unexpected error from a subscription resource function. So the `notifyFailure()` method in `SubscriptionCallback` class can be invoked. So that it can be covered by a test. 
I used it in `testSubscriptionFunctionWithErrors` test case

DimuthuMadushan (approved):

ThisaruGuruge (approved):

PR #691, opened 2022-04-19
Remove Unnecessary Usage of `var` Type

DimuthuMadushan:
## Purpose

Fixes: [#2844](https://github.com/ballerina-platform/ballerina-standard-library/issues/2844)

## Examples
N/A

## Checklist
- [x] Linked to an issue
- [ ] ~~Updated the changelog~~
- [ ] ~~Added tests~~
- [ ] ~~Updated the spec~~

### DIFF
--- a/ballerina/listener_utils.bal
+++ b/ballerina/listener_utils.bal
@@ -51,10 +51,10 @@ isolated function handlePostRequests(Engine engine, Context context, http:Reques
 
 isolated function getResponseFromJsonPayload(Engine engine, Context context, http:Request request,
                                              map<Upload|Upload[]> fileInfo = {}) returns http:Response {
-    var payload = request.getJsonPayload();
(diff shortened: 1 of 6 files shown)


ballerina/listener_utils.bal
```diff
@@ -51,10 +51,10 @@ isolated function handlePostRequests(Engine engine, Context context, http:Reques
 
 isolated function getResponseFromJsonPayload(Engine engine, Context context, http:Request request,
                                              map<Upload|Upload[]> fileInfo = {}) returns http:Response {
-    var payload = request.getJsonPayload();
+    json|http:ClientError payload = request.getJsonPayload();
     if payload is json {
```
ThisaruGuruge, line 55: Shall we change this logic a bit?

Check for the error first, return it, and then do other processing outside? We don't need to add the logic inside a block then.

ThisaruGuruge (approved):

PR #789, opened 2022-06-27
Improve WebSocket Service for GraphQL Subscriptions

Nuvindu:
## Purpose

Fixes [#2899](https://github.com/ballerina-platform/ballerina-standard-library/issues/2899)

## Checklist
- [ ] Linked to an issue
- [ ] Updated the changelog
- [ ] Added tests
- [ ] Updated the spec


ballerina-tests/tests/records.bal
```diff
@@ -229,3 +229,14 @@ public enum InstrumentType {
...
+# Represents the standard WebSocket message structure for subscriptions
+#
+# + 'type - Type of the WebSocket message  
+# + id - Operation id for uniquely identifying server responses and connecting them with the client's requests  
+# + payload - The requested data to go with the specific message type
+public type WSPayload record {|
+    string 'type;
+    string id?;
+    json payload?;
+|};
```
DimuthuMadushan, line 242: Can't we use the same `WSPayload` type defined in `records.bal` since it's a public type?

ballerina/websocket_service.bal
```diff
@@ -24,70 +25,111 @@ isolated service class WsService {
...
+                self.initiatedConnection = false;
             }
+            check sendWebSocketResponse(caller, self.customHeaders, WS_COMPLETE, null, connectionId);
+            closeConnection(caller);
+        } else if wsType.equalsIgnoreCaseAscii(WS_PING) {
+            json wsMessage = {"type": WS_PONG};
+            check caller->writeTextMessage(wsMessage.toString());
+        } else if wsType.equalsIgnoreCaseAscii(WS_PONG) {
+            json wsMessage = {"type": WS_PING};
+            check caller->writeTextMessage(wsMessage.toString());
```
DimuthuMadushan, line 130: It seems we can handle these two cases with [remote methods](https://github.com/ballerina-platform/module-ballerina-websocket/blob/main/docs/spec/spec.md#onping-and-onpong) associated with the WebSocket service. @Nuvindu can you check this as well?

Nuvindu: As I understand, `ping` and `pong` specified in the [graphql-ws spec](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#ping) are the types wrapped in text messages. So they should be handled in `onTextMessage()` function. Is that the correct approach?

ballerina/websocket_service.bal
```diff
@@ -24,70 +25,111 @@ isolated service class WsService {
...
+            parser:OperationNode|json node = validateSubscriptionPayload(wsPayload, self.engine);
+            if node is parser:OperationNode {
+                check executeOperation(self.engine, self.context, self.schema, self.customHeaders, caller,
+                                       connectionId, node);
             } else {
-                check caller->writeTextMessage(sourceStream.toJsonString());
+                check sendWebSocketResponse(caller, self.customHeaders, WS_ERROR, node, connectionId);
+                closeConnection(caller);
+            }
+        } else if (wsType.equalsIgnoreCaseAscii(WS_STOP) || wsType.equalsIgnoreCaseAscii(WS_COMPLETE)) {
```
DimuthuMadushan, line 110:
```suggestion
        } else if wsType.equalsIgnoreCaseAscii(WS_STOP) || wsType.equalsIgnoreCaseAscii(WS_COMPLETE) {
```

ballerina/types.bal
```diff
@@ -93,3 +93,14 @@ public type GenericResponseWithErrors record {|
...
+# Represents the standard WebSocket message structure for subscriptions
+#
+# + 'type - Type of the WebSocket message  
+# + id - Operation id for uniquely identifying server responses and connecting them with the client's requests  
+# + payload - The requested data to go with the specific message type
+public type WSPayload record {|
+    string 'type;
+    string id?;
+    json payload?;
+|};
```
DimuthuMadushan, line 106: What's the reason for making this public?

ballerina-tests/tests/records.bal
DimuthuMadushan: How did you try this? `graphql:WSPayload` should be worked.

ballerina/websocket_service.bal
DimuthuMadushan: In that case, this is fine.

ballerina-tests/tests/records.bal
Nuvindu: Ah. forgot to use it.

ballerina/types.bal
Nuvindu: No reason. I'll remove it.

DimuthuMadushan: If this is used only for testing purposes, no point in making this visible to the user.

Nuvindu: No. It is used in the build

docs/spec/spec.md
```diff
@@ -1258,6 +1274,12 @@ service on new graphql:Listener(4000) {
 ```
 > **Note:** The field enable accepts a `boolean` that denotes whether the client is enabled or not. By default, it has been set to `false`. The optional field `path` accepts a valid `string` for the GraphiQL service. If the path is not given in the configuration, `/graphiql` is set as the default path.
 
+#####  9.1.5.1 GraphiQL for Subscriptions
+GraphiQL uses WebSockets as the communication protocol for Subscription requests. And there is an underlying sub protocol called `graphql-transport-ws` which follows a standard message structure for subscriptions defined in the [specification](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md). Since GraphiQL follows the same sub protocol, all the subscription requests/responses will be wrapped in this format.
+
+> **Note:** There is another supported sub protocol named `graphql-ws`. GraphQL playground follows that sub protocol for subscriptions.  
+> If a subscription request occurs without a sub-protocol, responses will be dispatched according to the usual behavior. 
+
```
ThisaruGuruge, line 1282: This information about the protocols should come under the section `2.1 Listener`. There's not point of having them here.

ballerina-tests/tests/records.bal
```diff
@@ -229,3 +229,14 @@ public enum InstrumentType {
     # Instruments with leather, wooden or metal surfaces
     PERCUSSION
 }
+
+# Represents the standard WebSocket message structure for subscriptions
+#
+# + 'type - Type of the WebSocket message  
+# + id - Operation id for uniquely identifying server responses and connecting them with the client's requests  
+# + payload - The requested data to go with the specific message type
+public type WSPayload record {|
```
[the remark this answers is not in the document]
Nuvindu: it is not public now

native/src/main/resources/graphiql-subscriptions.html
```diff
@@ -0,0 +1,46 @@
...
+    </head>
+    <body style="margin: 0;"    >
+        <div id="graphiql" style="height: 100vh;"></div>
+
+        <script crossorigin src="https://unpkg.com/react/umd/react.production.min.js"></script>
+        <script crossorigin src="https://unpkg.com/react-dom/umd/react-dom.production.min.js"></script>
+        <script crossorigin src="https://unpkg.com/graphiql/graphiql.min.js"></script>
+
+        <script>
+                const fetcher = GraphiQL.createFetcher({ url: "${url}", subscriptionUrl: "${subscriptionUrl}" });
```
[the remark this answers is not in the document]
Nuvindu: There are two html files for `graphiql` now. One contains `subscriptionUrl` and other one doesn't. The one with `subscriptionUrl` is only called when the `Subscription` type is included in the schema.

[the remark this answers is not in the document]
ThisaruGuruge: Why do we need to maintain two HTML files here?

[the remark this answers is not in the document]
Nuvindu: Removed the other one

ballerina/constants.bal
```diff
@@ -55,3 +55,25 @@ const DECIMAL = "Decimal";
...
+const WS_COMPLETE = "complete";
+
+// WebSocket Sub Protocols
+const GRAPHQL_WS = "graphql-ws";
+const GRAPHQL_TRANSPORT_WS = "graphql-transport-ws";
+const WS_SUB_PROTOCOL = "Sec-WebSocket-Protocol";
+const DEFAULT_VALUE = "default";
+
+// Windows OS
+const WINDOWS = "WINDOWS_NT";
```
ThisaruGuruge, line 79: We don't need this now, right?

ballerina/listener.bal
```diff
@@ -54,7 +54,12 @@ public class Listener {
         if graphiql.enable {
             check validateGraphiqlPath(graphiql.path);
             string gqlServiceBasePath = name is () ? "" : getBasePath(name);
-            HttpService graphiqlService = getGraphiqlService(serviceConfig, gqlServiceBasePath);
+            Engine engine = check new (getSchemaString(serviceConfig), getMaxQueryDepth(serviceConfig));
```
ThisaruGuruge, line 57: Why do we need to create another instance of the `Engine`? This does not seem correct. We have to have only one Engine object.

ballerina/listener_utils.bal
```diff
@@ -368,8 +375,15 @@ isolated function getGraphiqlService(GraphqlServiceConfig? serviceConfig, string
         private final readonly & ListenerAuthConfig[]? authConfig = authConfigurations;
 
         isolated resource function get .(http:Caller caller) returns http:Response|http:InternalServerError {
-            string graphqlURL = string `http://${caller.localAddress.host}:${caller.localAddress.port}/${basePath}`;
-            string|error htmlAsString = getHtmlContentFromResources(graphqlURL);
+            string graphqlURL = os:getEnv("OS").toUpperAscii() == WINDOWS
```
ThisaruGuruge, line 378: Shall we update this?

ballerina/websocket_service.bal
```diff
@@ -24,70 +25,108 @@ isolated service class WsService {
...
+                check sendWebSocketResponse(caller, self.customHeaders, WS_ERROR, node, connectionId);
+                closeConnection(caller);
+            }
+        } else if wsType == WS_STOP || wsType == WS_COMPLETE {
+            lock {
+                string[] connections = [];
+                foreach string i in self.activeConnections {
+                    connections.push(i);
+                }
+                int? index = array:indexOf(connections, connectionId);
```
ThisaruGuruge, line 115: We don't need to import `lang.array` for this.

```suggestion
                int? index = connections.indexOf(connectionId);
```

docs/spec/spec.md
```diff
@@ -127,6 +127,8 @@ import ballerina/graphql;
 ### 2.1 Listener
 
 Since the GraphQL spec does not mandate an underlying client-server protocol, a GraphQL implementation can use any protocol underneath. The Ballerina GraphQL package, like most of the other implementations, uses HTTP as the protocol. The Ballerina GraphQL listener is using an HTTP listener to listen to incoming requests through HTTP.
+If the schema contains the `Subscription` type, The GraphQL listener will establish a new WebSocket listener to listen to incoming subscription requests.
+In Ballerina, WebSocket is used as the communication protocol for GraphQL subscriptions as it is capable of dispatching data continuously while maintaining a persistent connection. Additionally, Ballerina GraphQL supports `graphql-transport-ws` and `graphql-ws` websocket sub-protocols in subscriptions. If a WebSocket connection is established with one of these underlying sub-protocols, all the subscription responses will be wrapped in a standard message structure defined in their [specification](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md). If a subscription request occurs without a sub-protocol, responses will be dispatched according to the usual behavior.
```
ThisaruGuruge, line 131: What is "usual behavior"? Shall we describe that as well?

docs/spec/spec.md
```diff
@@ -294,14 +298,14 @@ As per the [GraphQL specification](https://spec.graphql.org/June2018/#sec-Mutati
 
 #### 3.1.3 The `Subscription` Type
 
-The `Subscription` type in a GraphQL schema is used to continuously fetch data from a GraphQL service. In Ballerina, the service itself is the schema, and each `resource` method with the `subscribe` accessor inside a GraphQL service is mapped to a field in the root `Subscription` type.
+`Subscription` is a root operation type in GraphQL schema but unlike queries, it is an optional type. In Ballerina, the service itself is the schema and all the subscription resolvers in the GraphQL service are mapped to fields in the root `Subscription` type.
```
ThisaruGuruge, line 301:
```suggestion
The `Subscription` type in a GraphQL schema is used to continuously fetch data from a GraphQL service. In Ballerina, each `resource` method with the `subscribe` accessor inside a GraphQL service is mapped to a field in the root `Subscription` type. 
```

native/src/main/java/io/ballerina/stdlib/graphql/runtime/engine/ListenerUtils.java
```diff
@@ -128,12 +128,19 @@ public static BString sanitizeBasePath(String basePath) {
...
     }
 
-    public static Object getHtmlContentFromResources(BString url) {
+    public static Object getHtmlContentFromResources(BString url, Object subscriptionUrl) {
         InputStream htmlAsStream = ClassLoader.getSystemResourceAsStream(GRAPHIQL_RESOURCE);
         try {
             byte[] bytes = htmlAsStream.readAllBytes();
             String htmlAsString = new String(bytes, StandardCharsets.UTF_8);
-            htmlAsString = htmlAsString.replace(REGEX_URL, url.getValue());
+            String graphiqlUrl;
```
ThisaruGuruge, line 136: Shall we use a StringBuilder for this?

ballerina/websocket_service.bal
```diff
@@ -24,70 +25,108 @@ isolated service class WsService {
...
-                check caller->writeTextMessage(sourceStream.toJsonString());
+                check sendWebSocketResponse(caller, self.customHeaders, WS_ERROR, node, connectionId);
+                closeConnection(caller);
+            }
+        } else if wsType == WS_STOP || wsType == WS_COMPLETE {
+            lock {
+                string[] connections = [];
+                foreach string i in self.activeConnections {
+                    connections.push(i);
+                }
```
ThisaruGuruge, line 114:
```suggestion
                string[] connections =  self.activeConnections.clone();
```

ballerina/websocket_service.bal
```diff
@@ -24,70 +25,108 @@ isolated service class WsService {
...
+                int? index = array:indexOf(connections, connectionId);
+                if index is int {
+                    _ = connections.remove(index);
+                }
```
ThisaruGuruge, line 118: What's the reason here to clone the `self.activeConnections` to a local variable and re-assigning it back?

ballerina/websocket_utils.bal
```diff
@@ -0,0 +1,112 @@
...
+        check sendWebSocketResponse(caller, customHeaders, WS_ERROR, sourceStream, connectionId);
+    }
+}
+
+isolated function validateSubscriptionPayload(json|WSPayload data, Engine engine) returns parser:OperationNode|json {
+    json|error payload = data is WSPayload ? data?.payload : data;
+    if payload is error {
+        return {errors: [{message: "Invalid format in WebSocket payload"}]};
+    }
+    json|error document = payload.query;
```
ThisaruGuruge, line 52: When `document is error`, what happens?

ballerina/websocket_utils.bal
```diff
@@ -0,0 +1,112 @@
...
+    if document !is string || document == "" {
+        return {errors: [{message: "Query not found"}]};
+    }
+    json|error variables = payload.variables is error ? () : payload.variables;
```
ThisaruGuruge, line 56: When `variables is error` what happens?

ballerina/websocket_utils.bal
```diff
@@ -0,0 +1,112 @@
...
+    return <stream<any, error?>>result;
+}
+
+isolated function sendWebSocketResponse(websocket:Caller caller, map<string> & readonly customHeaders, string wsType,
+                                        json payload, string? id = ()) returns websocket:Error? {
+    if !customHeaders.hasKey(WS_SUB_PROTOCOL) {
+        check caller->writeMessage(payload);
+    } else {
+        string 'type = getMessageType(wsType, customHeaders);
+        json jsonResponse = id != () ? {"type": 'type, id: id, payload: payload} : {"type": 'type, payload: payload};
```
ThisaruGuruge, line 83:
```suggestion
        json jsonResponse = id != () ? {'type: 'type, id: id, payload: payload} : {'type: 'type, payload: payload};
```

docs/spec/spec.md
Nuvindu: Sure

ballerina/websocket_service.bal
Nuvindu: `self.activeConnections` should be immutable here. Therefore I used `readonly` when the variable is initiated. But here I used a local variable to use the array as a mutable object.

ballerina/websocket_utils.bal
Nuvindu: It will return an error stating "Invalid format in request parameter: variables".

ballerina/websocket_service.bal
ThisaruGuruge: Any specific reason to make it readonly?

ballerina/websocket_service.bal
ThisaruGuruge: Why does it need to be immutable?

ballerina/websocket_utils.bal
ThisaruGuruge: In both cases, we are ignoring the error. It also reduces the readability of the code. We should handle this better.

ballerina/websocket_service.bal
Nuvindu: Unless `self.activeConnections` array gets empty during the runtime.

ballerina/websocket_utils.bal
```diff
@@ -0,0 +1,119 @@
...
+    return validation.toJson();
+}
+
+isolated function getSubscriptionResponse(Engine engine, __Schema schema, Context context,
+                                          parser:FieldNode node) returns stream<any, error?>|json {
+    ExecutorVisitor executor = new(engine, schema, context, {});
+    any|error result = getSubscriptionResult(executor, node);
+    if result !is stream<any, error?> {
+        return {errors: [{message: "Error/s occurred in the subscription resolver"}]};
+    }
```
DimuthuMadushan, line 80: @Nuvindu Can't the subscription resolvers return `error` values? If the resolvers can produce errors, the error message will not be appeared in the response due to this logic. The error message is always overridden with the `Error/s occurred in the subscription resolver`.

Nuvindu: Yes. I should bind the error message to the Websocket payload. Missed it.

ballerina/websocket_utils.bal
```diff
@@ -0,0 +1,123 @@
...
+        check caller->writeMessage(payload);
+    } else {
+        string 'type = getMessageType(wsType, customHeaders);
+        json jsonResponse = id != () ? {'type: 'type, id: id, payload: payload} : {'type: 'type, payload: payload};
+        check caller->writeMessage(jsonResponse);
+    }
+}
+
+isolated function getMessageType(string wsType, map<string> & readonly customHeaders) returns string {
+    if customHeaders.get(WS_SUB_PROTOCOL) == GRAPHQL_WS {
```
ThisaruGuruge, line 100: This will panic if the `WS_SUB_PROTOCOL` is not found in the map. Should first use the `hasKey` function to check whether the key is present.

docs/spec/spec.md
```diff
@@ -135,6 +135,8 @@ import ballerina/graphql;
 ### 2.1 Listener
 
 Since the GraphQL spec does not mandate an underlying client-server protocol, a GraphQL implementation can use any protocol underneath. The Ballerina GraphQL package, like most of the other implementations, uses HTTP as the protocol. The Ballerina GraphQL listener is using an HTTP listener to listen to incoming requests through HTTP.
+If the schema contains the `Subscription` type, The GraphQL listener will establish a new WebSocket listener to listen to incoming subscription requests.
```
ThisaruGuruge, line 138: Let's add two sub-topics for `2.1.1 HTTP Listener` and `2.1.2 WebSocket Listener`.

docs/spec/spec.md
```diff
@@ -135,6 +135,8 @@ import ballerina/graphql;
...
+In Ballerina, WebSocket is used as the communication protocol for GraphQL subscriptions as it is capable of dispatching data continuously while maintaining a persistent connection. Additionally, Ballerina GraphQL supports `graphql-transport-ws` and `graphql-ws` websocket sub-protocols in subscriptions. If a WebSocket connection is established with one of these underlying sub-protocols, all the subscription responses will be wrapped in a standard message structure defined in their [specification](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md). In a standard response, it includes JSON fields for `type`, `id`, and `payload`. The `type` field specifies the message type of the response. And the `id` field is used to uniquely identify the client. And the `payload` field includes the GraphQL response returned from the GraphQL engine. If a subscription request occurs without a sub-protocol, only the GraphQL response will be dispatched. All the responses will be converted to text before sending them to the clients.
```
ThisaruGuruge, line 139: `All the responses will be converted to text before sending them to the clients` Is this true? Aren't we sending the responses as JSON?

docs/spec/spec.md
```diff
@@ -302,14 +306,14 @@ As per the [GraphQL specification](https://spec.graphql.org/June2018/#sec-Mutati
 
 #### 3.1.3 The `Subscription` Type
 
-The `Subscription` type in a GraphQL schema is used to continuously fetch data from a GraphQL service. In Ballerina, the service itself is the schema, and each `resource` method with the `subscribe` accessor inside a GraphQL service is mapped to a field in the root `Subscription` type.
+The `Subscription` type in a GraphQL schema is used to continuously fetch data from a GraphQL service. In Ballerina, each `resource` method with the `subscribe` accessor inside a GraphQL service is mapped to a field in the root `Subscription` type.
```
ThisaruGuruge, line 309:
```suggestion
The `Subscription` type in a GraphQL schema is used to continuously fetch data from a GraphQL service. In Ballerina, each `resource` method with the `subscribe` accessor inside a GraphQL service is mapped to a field in the root `Subscription` type.
```

ballerina/websocket_utils.bal
Nuvindu: It is already checked before the `getMessageType` method is called.

native/src/main/java/io/ballerina/stdlib/graphql/runtime/engine/ListenerUtils.java
```diff
@@ -128,12 +128,18 @@ public static BString sanitizeBasePath(String basePath) {
...
-    public static Object getHtmlContentFromResources(BString url) {
+    public static Object getHtmlContentFromResources(BString url, Object subscriptionUrl) {
         InputStream htmlAsStream = ClassLoader.getSystemResourceAsStream(GRAPHIQL_RESOURCE);
         try {
             byte[] bytes = htmlAsStream.readAllBytes();
             String htmlAsString = new String(bytes, StandardCharsets.UTF_8);
-            htmlAsString = htmlAsString.replace(REGEX_URL, url.getValue());
+            StringBuilder graphiqlUrl = new StringBuilder("{ url: \"" + url.getValue() + "\" }");
+            if (subscriptionUrl != null) {
+                graphiqlUrl.deleteCharAt(graphiqlUrl.length() - 1);
```
ThisaruGuruge, line 138: Instead of adding and deleting the character, you can add the close brace at the end.

ThisaruGuruge (approved):

PR #812, opened 2022-07-14
Introduce GraphQL Interceptors

DimuthuMadushan:
## Purpose
This PR consists of GraphQL `Interceptor` implementation. Currently, the Interceptors are supported only for `Query` and `Mutation` operations.

Fixes: [#2001](https://github.com/ballerina-platform/ballerina-standard-library/issues/2001)

## Examples
Consider the following GraphQL service. It includes two service level interceptors named `ServiceInterceptor` and `Scope`.
```ballerina
import ballerina/graphql;
import ballerina/log;

readonly service class ServiceInterceptor {
    *graphql:Interceptor;
    remote function execute(graphql:Context ctx, graphql:Field 'field) returns anydata|error {
        log:printInfo(string `Service Interceptor execution!`);
        var result = ctx.resolve('field);
        log:printInfo("Connection closed!");
        return result;
    }
}
 
service class Scope {
    *graphql:Interceptor;
    remote function execute(graphql:Context ctx, graphql:Field 'field) returns anydata|error {
        log:printInfo("Execution Scope: Admin");
        var result = ctx.resolve('field);
        log:printInfo("Leaving Admin Scope!");
        return result;
    }
}
 
@graphql:ServiceConfig {
   interceptors: [new ServiceInterceptor(), new Scope()]
}
service /graphql on new graphql:Listener(9000) {
 
    resource function get name(int id) returns string {
        log:printInfo("Resolver: name");
        return "Ballerina";
    }
}
```
If this service is queried as follows:
```graphql
query {
    name
}
```
It will log the statements as follow:
```shell
1. Service Interceptor execution!
2. Execution Scope: Admin
3. Resolver: name
4. Leaving Admin Scope!
5. Connection closed!
```

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [ ] ~Updated the spec~


ballerina/annotations.bal
```diff
@@ -22,13 +22,15 @@
...
 # + schemaString - The generated schema. This is auto-generated at the compile-time
+# + interceptors - GraphQL service level interceptors
 public type GraphqlServiceConfig record {|
     int maxQueryDepth?;
     ListenerAuthConfig[] auth?;
     ContextInit contextInit = initDefaultContext;
     CorsConfig cors?;
     Graphiql graphiql = {};
     readonly string schemaString = "";
+    (readonly & Interceptor)[] interceptors = [];
```
ThisaruGuruge, line 33: SHouldn't this be:

```suggestion
    readonly (readonly & Interceptor)[] interceptors = [];
```

ballerina-tests/tests/interceptors.bal
```diff
@@ -0,0 +1,284 @@
...
+        }
+        return result;
+    }
+}
+
+readonly service class InvalidInterceptor7 {
+    *graphql:Interceptor;
+
+    isolated remote function execute(graphql:Context ctx, graphql:Field 'field) returns anydata|error {
+        var result = ctx.resolve();
```
ThisaruGuruge, line 269: Every interceptor here calls the `ctx.resolve()` in the first line and then processes it. We should have a test to execute something before, then call `resolve` and then do another thing. We need to test this functionality whether it executes in the same order as intended.

ballerina/context.bal
```diff
@@ -97,6 +99,35 @@ public isolated class Context {
     isolated function getFileInfo() returns map<Upload|Upload[]> = @java:Method {
         'class: "io.ballerina.stdlib.graphql.runtime.engine.EngineUtils"
     } external;
+
+    public isolated function resolve() returns anydata|error {
```
ThisaruGuruge, line 103: The API should be `resolve(Field 'field) returns anydata`. It will return null if the field resolution returns an error.

ballerina/engine.bal
```diff
@@ -21,20 +21,40 @@ import graphql.parser;
...
+    private int nextInterceptor;
 
-    isolated function init(string schemaString, int? maxQueryDepth, Service s) returns Error? {
+    isolated function init(string schemaString, int? maxQueryDepth, Service s, (readonly & Interceptor)[] interceptors) returns Error? {
         if maxQueryDepth is int && maxQueryDepth < 1 {
             return error Error("Max query depth value must be a positive integer");
         }
         self.maxQueryDepth = maxQueryDepth;
         self.schema = check createSchema(schemaString);
+        self.interceptors = interceptors.cloneReadOnly();
```
ThisaruGuruge, line 33: If we mark the `interceptors` field as readonly, we don't need to use `cloneReadOnly()` here.

ballerina/engine.bal
```diff
@@ -21,20 +21,40 @@ import graphql.parser;
...
+    }
+
+    isolated function getNextInterceptor() returns (readonly & Interceptor)? {
+        lock {
+            if self.interceptors.length() > self.nextInterceptor {
+                readonly & Interceptor next = self.interceptors[self.nextInterceptor];
+                self.nextInterceptor += 1;
+                return next;
+            }
+            self.nextInterceptor = 0;
```
ThisaruGuruge, line 53: How does this work parallelly? If there are two requests processed parallelly, and one is processed the first interceptor, while the other has processed the third interceptor, what will be the result of this function?

The engine should not be used to store request-related data. The context should be used for that.

ballerina/executor_visitor.bal
```diff
@@ -90,6 +82,30 @@ class ExecutorVisitor {
...
+    isolated function executeMutation(parser:FieldNode fieldNode, parser:RootOperationType operationType) {
+        (string|int)[] path = [fieldNode.getName()];
+        string operationTypeName = getOperationTypeNameFromOperationType(operationType);
+        __Type parentType = <__Type>getTypeFromTypeArray(self.schema.types, operationTypeName);
+        __Type fieldType = getFieldTypeFromParentType(parentType, self.schema.types, fieldNode);
+        Field 'field = new (fieldNode, self.engine.getService(), fieldType, path, operationType);
+        self.context.setField('field);
+        var result = self.engine.resolve(self.context, 'field);
+        self.errors = self.context.getErrors();
+        self.data[fieldNode.getAlias()] = result;
```
ThisaruGuruge, line 106:
```suggestion
        self.data[fieldNode.getAlias()] = result is ErrorDetail ? () : result;
```

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/InterceptorAnalysisTask.java
```diff
@@ -0,0 +1,44 @@
...
+            return;
+        }
+        ClassDefinitionNode classDefinitionNode = (ClassDefinitionNode) context.node();
+        NodeList<Token> tokens = classDefinitionNode.classTypeQualifiers();
+        if (tokens.isEmpty()) {
+            return;
+        }
+        if (!tokens.stream().allMatch(token -> READONLY_SERVICE.contains(token.text()))) {
+            return;
+        }
```
ThisaruGuruge, line 30: This should be validated using the Semantic API. Syntax API validations are not trivial.

In this case, consider if a service is defined using `readonly service` with more than one space in between. Or even worse, `readonly` in one line, and the `service` is in the next line. We cannot check it correctly. But using the Semantics APIs, we can check the qualifiers of a service definition, so the syntax does not matter anymore. 

Check how the validation is done for a GraphQL service.

ballerina/annotations.bal
DimuthuMadushan: Added

ballerina-tests/tests/interceptors.bal
DimuthuMadushan: Added a test.

ballerina/context.bal
```diff
@@ -97,6 +99,35 @@ public isolated class Context {
...
+        if self.getEngine() is Engine {
+            Engine engine = <Engine>self.getEngine();
+            Field 'field = <Field>self.getField(self);
```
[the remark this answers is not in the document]
DimuthuMadushan: Removed.

ballerina/engine.bal
DimuthuMadushan: Updated

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/service/validator/InterceptorValidator.java
```diff
@@ -0,0 +1,87 @@
...
+
+    public void validate() {
+        NodeList<Node> members = this.classDefinitionNode.members();
+        for (Node member : members) {
+            if (member.kind() == SyntaxKind.TYPE_REFERENCE) {
+                String typeReference = ((TypeReferenceNode) member).typeName().toString();
+                if (typeReference.equals(GRAPHQL_INTERCEPTOR)) {
+                    validateInterceptorService(members);
+                    break;
+                }
```
ThisaruGuruge, line 43: This is not accurate to check the type name. Check [`isFileUploadParameter()`](https://github.com/ballerina-platform/module-ballerina-graphql/blob/3a7dbfb6ee057d95f91f5b9dbf386f7e32dffbf5/compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/Utils.java#L204) method for example.

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/service/validator/InterceptorValidator.java
```diff
@@ -0,0 +1,98 @@
...
+    public boolean isInterceptor(Node member) {
+        if (this.context.semanticModel().symbol(member).isEmpty()) {
+            return false;
+        }
+        Symbol symbol = this.context.semanticModel().symbol(member).get();
+        TypeReferenceTypeSymbol typeReferenceTypeSymbol = (TypeReferenceTypeSymbol) symbol;
+        if (typeReferenceTypeSymbol.getName().isEmpty()) {
+            return false;
+        }
+        return GRAPHQL_INTERCEPTOR.equals(typeReferenceTypeSymbol.getName().get());
```
ThisaruGuruge, line 91: This does not check the package name and the org, right?

DimuthuMadushan: Missed that. Fixed in https://github.com/ballerina-platform/module-ballerina-graphql/pull/812/commits/0ccd32090205c2201acc05c687830c44b0ce4a2a

ThisaruGuruge (approved):

PR #834, opened 2022-07-22
Fix Not Initializing Context per Request in GraphQL Subscriptions

DimuthuMadushan:
## Purpose

Fixes: [#3115](https://github.com/ballerina-platform/ballerina-standard-library/issues/3115)

## Examples
N/A

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [ ] ~Updated the spec~


ballerina/websocket_service.bal
```diff
@@ -25,10 +25,10 @@ isolated service class WsService {
     private final readonly & __Schema schema;
     private final Context context;
 
-    isolated function init(Engine engine, __Schema & readonly schema) {
+    isolated function init(Engine engine, __Schema & readonly schema, Context context) {
         self.engine = engine;
         self.schema = schema;
-        self.context = new;
+        self.context = context;
```
ThisaruGuruge, line 31: Why do we need to store the context in the service? It does not make any sense.

changelog.md
```diff
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Fixed
 - [[#3069] Fix Enums with String Values Return String Value for Enum Name](https://github.com/ballerina-platform/ballerina-standard-library/issues/3069)
 - [[#3067] Fix Single Quote Character Included in Field and Argument Names](https://github.com/ballerina-platform/ballerina-standard-library/issues/3067)
+- [[#3115] Fix Not Initializing Context per Service in GraphQL Subscriptions](https://github.com/ballerina-platform/ballerina-standard-library/issues/3115)
```
ThisaruGuruge, line 16: This gives an entirely different message about the fix.

Not initializing Context per service is the way to go. There's nothing to fix in it. In fact, the issue is initializing context per service.

ThisaruGuruge (changes requested):

ballerina/websocket_service.bal
DimuthuMadushan: AFAIK, When a user initializes a WebSocket connection, only the first request goes through the Upgrade Service. Then we can access the `http: Request` and create the context object. After the initialization of the connection, other requests go through the WebSocket remote functions. Since context can not be created after the initial request, store the context here to use for the rest of the messages in that connection.

ThisaruGuruge: That is wrong. The WebSocket service is initialized once for the listener. How are you going to handle multiple requests with this approach? 
A context is a request-specific object. You cannot create it per service, or cannot use a single context object for multiple requests. That's the whole point of the issue you are trying to fix here.

ThisaruGuruge (approved):

PR #858, opened 2022-08-05
Fix GraphQL Interceptors Skipping Fields when Returning Errors

DimuthuMadushan:
## Purpose

Fixes: [#3206](https://github.com/ballerina-platform/ballerina-standard-library/issues/3206)
Fixes: [#3221](https://github.com/ballerina-platform/ballerina-standard-library/issues/3221)

## Examples
N/A

## Checklist
- [x] Linked to an issue
- [ ] ~Updated the changelog~
- [x] Added tests
- [ ] ~Updated the spec~


ballerina/context.bal
```diff
@@ -132,10 +132,16 @@ public isolated class Context {
...
+            self.resetInterceptorCount();
             return;
         }
     }
+
+    isolated function resetInterceptorCount() {
+        lock {
+            self.nextInterceptor = 0;
+        }
+    }
```
ThisaruGuruge, line 144: Will this solve the issue correctly? What happens when the parallel field execution?
Maybe we need to think about a way to handle this more elegantly.

DimuthuMadushan: Earlier the problem was not resetting the interceptor counter when that interceptor is returning an error without executing the actual resolver. Don't we run fields sequentially for a request on the ballerina side? If so, since context is being created for every request, this won't be an issue. will it?

ThisaruGuruge (approved):

PR #907, opened 2022-09-06
Add schema generation support for module level graphql:Service variable declaration

MohamedSabthar:
## Purpose

Fixes: [#3294](https://github.com/ballerina-platform/ballerina-standard-library/issues/3294)

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [ ] ~Updated the spec~


compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/GraphqlCodeModifier.java
```diff
@@ -42,12 +42,13 @@ public GraphqlCodeModifier() {
 
     @Override
     public void init(CodeModifierContext modifierContext) {
-        modifierContext.addSyntaxNodeAnalysisTask(new ServiceAnalysisTask(this.modifierContextMap),
-                                                  SyntaxKind.SERVICE_DECLARATION);
+        modifierContext.addSyntaxNodeAnalysisTask(new ServiceDeclarationAnalysisTask(this.modifierContextMap),
+                SyntaxKind.SERVICE_DECLARATION);
```
ThisaruGuruge, line 46: Let's keep the indentation consistent with the other places. Better if you can update the IDE settings to use the existing indentation patterns.

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/service/validator/ServiceValidator.java
```diff
@@ -299,7 +323,7 @@ private void validateReturnType(TypeSymbol typeSymbol, Location location) {
                 break;
             case RECORD:
                 addDiagnostic(CompilationError.INVALID_ANONYMOUS_FIELD_TYPE, location, typeSymbol.signature(),
-                              getCurrentFieldPath());
```
ThisaruGuruge, line 302: Let's not change the existing indentations.

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/Utils.java
```diff
@@ -243,4 +252,26 @@ public static boolean isContextParameter(TypeSymbol typeSymbol) {
     public static String getAccessor(ResourceMethodSymbol resourceMethodSymbol) {
         return resourceMethodSymbol.getName().orElse(null);
     }
+
+    public static boolean isGraphQLServiceObjectDeclaration(
+            ModuleVariableDeclarationNode moduleVariableDeclarationNode) {
```
ThisaruGuruge, line 257: Let's update the formatting here.

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/ModuleLevelVariableDeclarationAnalysisTask.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+        DocumentId documentId = context.documentId();
+        String description = null;
+        if (context.semanticModel().symbol(moduleVariableDeclarationNode).isPresent()) {
+            VariableSymbol serviceVariableSymbol = (VariableSymbol) context.semanticModel()
+                    .symbol(moduleVariableDeclarationNode).get();
+            description = getDescription(serviceVariableSymbol);
+        }
+        Schema schema = generateSchema(interfaceFinder, graphqlServiceObjectNode, description, context);
+        addToModifierContextMap(documentId, moduleVariableDeclarationNode, schema);
```
ThisaruGuruge, line 81: These seem to be similar to the other analysis task code. Can we use a single method to call from both places instead of duplicating the code?

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/schema/generator/GraphqlSourceModifier.java
```diff
@@ -151,12 +188,11 @@ private AnnotationNode updateAnnotationNode(AnnotationNode annotationNode, Strin
...
-                    getUpdatedFields(annotationNode.annotValue().get(), schemaString);
-            MappingConstructorExpressionNode node =
-                    NodeFactory.createMappingConstructorExpressionNode(
-                            NodeFactory.createToken(SyntaxKind.OPEN_BRACE_TOKEN), updatedFields,
-                            NodeFactory.createToken(SyntaxKind.CLOSE_BRACE_TOKEN));
+            SeparatedNodeList<MappingFieldNode> updatedFields = getUpdatedFields(annotationNode.annotValue().get(),
+                    schemaString);
+            MappingConstructorExpressionNode node = NodeFactory.createMappingConstructorExpressionNode(
+                    NodeFactory.createToken(SyntaxKind.OPEN_BRACE_TOKEN), updatedFields,
+                    NodeFactory.createToken(SyntaxKind.CLOSE_BRACE_TOKEN));
```
ThisaruGuruge, line 195: When sending a PR, let's check whether we have changed lines that are not relevant to the PR, and try not to change them.

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/schema/generator/SchemaGenerator.java
```diff
@@ -152,6 +164,35 @@ private void findRootTypes(ServiceDeclarationSymbol serviceDeclarationSymbol) {
...
+
+        return new ArrayList<>();
+    }
+
+    private Collection<? extends MethodSymbol> getMethods(
+            ObjectConstructorExpressionNode objectConstructorExpressionNode) {
+        // noinspection OptionalGetWithoutIsPresent
+        return objectConstructorExpressionNode.members().stream()
+                .filter(member -> member instanceof FunctionDefinitionNode && context.semanticModel().symbol(member)
+                        .isPresent()).map(methodNode -> (MethodSymbol) context.semanticModel().symbol(methodNode).get())
```
ThisaruGuruge, line 184: Let's avoid using `instanceof` checks as much as possible. The SymbolKind and TypeKind are there to avoid these checks and to help downcasting.

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/Utils.java
MohamedSabthar: This line exceeds the max char limit if aligned with the bracket

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/schema/generator/GraphqlSourceModifier.java
MohamedSabthar: restored to previous

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/GraphqlCodeModifier.java
MohamedSabthar: Updated

ThisaruGuruge:
Let's follow the best practices in commit messages and pull requests.
- Check the PR title, and make sure it conveys the message correctly. Try to answer the question, "What does this PR do?".
- It's better to have a small description for a PR unless the fix is an obvious one.
- Adding examples too can help if the PR is more complex.
- Go through the PR checklist always, and make sure to tick all the boxes that are covered, and ~strikethrough~ the ones that don't apply for the particular PR.
- If you cannot check an item in the checklist as checked and cannot strikethrough it either, you're probably missing something in the PR.

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/schema/generator/SchemaGenerator.java
MohamedSabthar: Updated

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/ModuleLevelVariableDeclarationAnalysisTask.java
MohamedSabthar: Common implementations are moved to an abstract class

ballerina-tests/tests/01_listener_test.bal
```diff
@@ -26,3 +27,59 @@ function testInvalidMaxQueryDepth() returns error? {
     graphql:Error err = <graphql:Error>result;
     test:assertEquals(err.message(), "Max query depth value must be a positive integer");
 }
+
+graphql:Service greetingService = service object {
+    resource function get greeting() returns string {
+        return "Hello, World";
+    }
+};
```
ThisaruGuruge, line 35: Let's move the services to the test_services file.

ballerina-tests/tests/01_listener_test.bal
```diff
@@ -26,3 +27,59 @@ function testInvalidMaxQueryDepth() returns error? {
...
+    };
+    runtime:deregisterListener(serverEP);
+    test:assertEquals(actualPayload, expectedPayload);
+}
+
+graphql:Service greetingService2 = @graphql:ServiceConfig {maxQueryDepth: 5} service object {
+    resource function get greeting() returns string {
+        return "Hello, World";
+    }
+};
```
ThisaruGuruge, line 63: Ditto

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/ModuleLevelVariableDeclarationAnalysisTask.java
```diff
@@ -0,0 +1,84 @@
...
+        super(nodeMap);
+    }
+
+    @Override
+    public void perform(SyntaxNodeAnalysisContext context) {
+        if (hasCompilationErrors(context)) {
+            return;
+        }
+
+        ModuleVariableDeclarationNode moduleVariableDeclarationNode = (ModuleVariableDeclarationNode) context.node();
```
ThisaruGuruge, line 52: Let's add a kind check before casting, just to be in the safe side.

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/ModuleLevelVariableDeclarationAnalysisTask.java
```diff
@@ -0,0 +1,84 @@
...
+        if (!isGraphQLServiceObjectDeclaration(moduleVariableDeclarationNode)) {
+            return;
+        }
+
+        if (moduleVariableDeclarationNode.initializer().isEmpty()) {
+            return;
```
ThisaruGuruge, line 58: Is there a way to go to the variable assignment from the variable declaration node? If so, we can handle this case as well, right?

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/ModuleLevelVariableDeclarationAnalysisTask.java
```diff
@@ -0,0 +1,84 @@
...
+        }
+        ObjectConstructorExpressionNode graphqlServiceObjectNode
+                = (ObjectConstructorExpressionNode) moduleVariableDeclarationNode.initializer().get();
```
ThisaruGuruge, line 61: Is it guaranteed that this will be an ObjectConstructorNode? I think it is better to check before casting.

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/ModuleLevelVariableDeclarationAnalysisTask.java
```diff
@@ -0,0 +1,84 @@
...
+        ServiceValidator serviceObjectValidator = getServiceValidator(context, graphqlServiceObjectNode,
+                                                                      interfaceFinder);
+        if (serviceObjectValidator.isErrorOccurred()) {
+            return;
+        }
+
+        String description = getDescription(context, moduleVariableDeclarationNode);
+        Schema schema = generateSchema(context, interfaceFinder, graphqlServiceObjectNode, description);
+        DocumentId documentId = context.documentId();
+        addToModifierContextMap(documentId, moduleVariableDeclarationNode, schema);
```
ThisaruGuruge, line 72: Can't we remove the duplication here with the abstract classes? Since the analysis tasks behave similarly, we can pass the corresponding analysis task and then check for errors, find interfaces, generate schema, etc.

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/Utils.java
ThisaruGuruge: This is a subjective take on this, but I think the variable name is too long. We can use something like `variableNode` here, since it is clear in this context. We can reduce the characters that way. Note, the intention here is not to change the name for the sake of shortening the line.

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/Utils.java
```diff
@@ -251,4 +261,31 @@ public static boolean isContextParameter(TypeSymbol typeSymbol) {
...
+        TypedBindingPatternNode typedBindingPatternNode = moduleVariableDeclarationNode.typedBindingPattern();
+        TypeDescriptorNode typeDescriptorNode = typedBindingPatternNode.typeDescriptor();
+        if (typeDescriptorNode.kind() != SyntaxKind.QUALIFIED_NAME_REFERENCE) {
+            return false;
+        }
+        return isGraphqlServiceQualifiedNameReference((QualifiedNameReferenceNode) typeDescriptorNode);
```
ThisaruGuruge, line 272: Check the casting here as well.

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/Utils.java
```diff
@@ -251,4 +261,31 @@ public static boolean isContextParameter(TypeSymbol typeSymbol) {
...
+        if (!PACKAGE_NAME.equals(modulePrefixToken.text())) {
+            return false;
+        }
+        IdentifierToken identifier = nameReferenceNode.identifier();
+        return SERVICE_NAME.equals(identifier.text());
+    }
+
+    public static boolean isFunctionDefinition(Node node) {
+        return node.kind() == SyntaxKind.RESOURCE_ACCESSOR_DEFINITION
+                || node.kind() == SyntaxKind.OBJECT_METHOD_DEFINITION;
```
ThisaruGuruge, line 289: What's the purpose of this method? The name and the logic don't seem to be matching.

ThisaruGuruge:
Let's update the PR title and the description.

ballerina-tests/tests/01_listener_test.bal
```diff
@@ -26,3 +27,59 @@ function testInvalidMaxQueryDepth() returns error? {
...
+
+@test:Config {
+    groups: ["listener", "client"]
+}
+function testAttachingGraphQLServiceToDynamicListener() returns error? {
+    graphql:Listener serverEP = check new (9600);
```
DimuthuMadushan, line 41: Let's use an existing listener in `test_listeners.bal` if possible.

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/ModuleLevelVariableDeclarationAnalysisTask.java
MohamedSabthar: This task is registered to run only when encountering SyntaxKind.MODULE_VAR_DECL. If check is redundant here

ballerina-tests/tests/01_listener_test.bal
MohamedSabthar: Updated

ballerina-tests/tests/01_listener_test.bal
MohamedSabthar: Moved

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/ModuleLevelVariableDeclarationAnalysisTask.java
MohamedSabthar: Updated

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/Utils.java
MohamedSabthar: Purpose of this method is to check provided node is a FunctionDefinitionNode in an object

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/ModuleLevelVariableDeclarationAnalysisTask.java
ThisaruGuruge: It is not redundant as per the compiler team. They specifically mentioned this in their demo about the compiler plugins.

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/Utils.java
ThisaruGuruge: But this returns whether it is a resource accessor definition or an object method definition?

MohamedSabthar: resource accessor definition and object method definition are the only 2 method definition we could have in an object right? both are instances of FunctionDefinitionNode.

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/ModuleLevelVariableDeclarationAnalysisTask.java
MohamedSabthar: Updated

MohamedSabthar:
> Let's update the PR title and description. Let's also not remove the checkbox when striking through an item.

Updated

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/ModuleLevelVariableDeclarationAnalysisTask.java
MohamedSabthar: It's not possible to go from variable declaration node but could do other way around. In that case we could handle this there

ballerina-tests/tests/01_listener_test.bal
MohamedSabthar: Updated

ThisaruGuruge (approved):

PR #936, opened 2022-09-19
Improve Compilation Error Messages to be more specific

MohamedSabthar:
## Purpose

Fixes: [#3062](https://github.com/ballerina-platform/ballerina-standard-library/issues/3062)

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [ ] ~~Updated the spec~~


compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/service/errors/ErrorMessage.java
```diff
@@ -26,44 +26,50 @@
...
-    ERROR_134("Anonymous record `{0}` cannot be used as an input object type of the field `{1}`");
+    ERROR_123("Non-distinct service class `{0}` is used as a GraphQL interface implementation"),
+    ERROR_124("Invalid hierarchical resource path `{0}` found in subscribe resource"),
+    ERROR_125("Invalid return type `{0}` found in subscribe resource `{1}`. GraphQL subscribe resource must return "
+                      + "`stream` type"),
+    ERROR_126("Invalid GraphQL resource accessor `{0}` found in resource `{1}`. Only " + RESOURCE_FUNCTION_GET +
+                      " and " + RESOURCE_FUNCTION_SUBSCRIBE + " are allowed"),
+    ERROR_127("Failed to generate the schema from the service. {0}"),
+    ERROR_128("Invalid resource method `{0}` found in GraphQL interceptor. GraphQL interceptors can not have resource "
+                      + "functions"),
```
ThisaruGuruge, line 68: Let's make the wording consistent. The terms `resource method` and `resource function` both are used in the same message. Let's stick to one term and use it every place. I'd suggest we use the `resource method` and `remote method`, to keep consistency with the other documentation.

ThisaruGuruge (approved):

PR #941, opened 2022-09-22
Fix Errors Not Identifying as Invalid Input Types

ThisaruGuruge:
## Purpose

$Subject

With this PR, the compiler plugin test directories are renamed for better readability. The package names are also updated to `test_package` so that the `Ballerina.toml` looks the same for every test package.

Fixes: [#3355](https://github.com/ballerina-platform/ballerina-standard-library/issues/3355)

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [x] Updated the spec


MohamedSabthar (approved):

docs/spec/spec.md
```diff
@@ -682,9 +684,40 @@ enum Direction {
...
 
-### 4.5 Input Objects
+### 4.5 Input Types
 
-In GraphQL, a field can have zero or more input arguments. These arguments can be either `Scalar` type, `Enum` type, or `Object` type. Although `Scalar` and `enum` types can be used as input and output types without a limitation, an object type can not be used as an input type and an output type. Therefore, separate kinds of objects are used to define input objects.
+In GraphQL, a field can have zero or more input arguments. These arguments can be either `Scalar` type, `Enum` type, or `Object` type.
+
+#### 4.5.1 Input Union Types
+
+An input type can be a Ballerina union type, if and only if the union consists of one of the supported types and the other member type is `nil`. A union with nil means the input type is `NON_NULL`. Any other union type will be resulting in a compilation error.
```
MohamedSabthar, line 693: Is `NON_NULL` correct here ? isn't it the other type in the union ?
ex: for input `string? name` isn't it the type is SCALAR (String)

ThisaruGuruge, line 693:
```suggestion
An input type can be a Ballerina union type, if and only if the union consists of one of the supported types and the other member type is `nil`. A union with nil means the input type is `String`. Any other union type will be resulting in a compilation error.
```

ThisaruGuruge: Fixed.

docs/spec/spec.md
```diff
@@ -682,9 +684,40 @@ enum Direction {
...
 
-### 4.5 Input Objects
+### 4.5 Input Types
 
-In GraphQL, a field can have zero or more input arguments. These arguments can be either `Scalar` type, `Enum` type, or `Object` type. Although `Scalar` and `enum` types can be used as input and output types without a limitation, an object type can not be used as an input type and an output type. Therefore, separate kinds of objects are used to define input objects.
+In GraphQL, a field can have zero or more input arguments. These arguments can be either `Scalar` type, `Enum` type, or `Object` type.
+
+#### 4.5.1 Input Union Types
+
+An input type can be a Ballerina union type, if and only if the union consists of one of the supported types and the other member type is `nil`. A union with nil means the input type is `String`. Any other union type will be resulting in a compilation error.
```
ThisaruGuruge, line 693:
```suggestion
An input type can be a Ballerina union type, if and only if the union consists of one of the supported types and the other member type is `nil`. A union with nil means the input type is a nullable type. Any other union type will be resulting in a compilation error.
```

PR #953, opened 2022-09-29
Introduce GraphQL Client Configuration

MohamedSabthar:
## Purpose

New graphql:ClientConfiguration introduced as a included record argument in client init method

Fixes: [#3431](https://github.com/ballerina-platform/ballerina-standard-library/issues/3431), [#3384](https://github.com/ballerina-platform/ballerina-standard-library/issues/3384)

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [x] Updated the spec


ballerina/types.bal
```diff
@@ -93,3 +93,38 @@ public type GenericResponseWithErrors record {|
...
+    http:PoolConfiguration? poolConfig = ();
+    http:CacheConfig cache = {};
+    http:Compression compression = http:COMPRESSION_AUTO;
+    http:ClientAuthConfig? auth = ();
+    http:CircuitBreakerConfig? circuitBreaker = ();
+    http:RetryConfig? retryConfig = ();
+    http:CookieConfig? cookieConfig = ();
+    http:ResponseLimitConfigs responseLimits = {};
+    http:ClientSecureSocket? secureSocket = ();
+    http:ProxyConfig? proxy = ();
```
ThisaruGuruge, line 128: We should introduce these types in GraphQL package as well.

ThisaruGuruge (commented):
Let's add a test for adding these cofigs

ballerina/types.bal
MohamedSabthar: updated

MohamedSabthar:
> Let's add a test for adding these cofigs

A basic auth test added

ThisaruGuruge (commented):
Let's add the spec update in this PR as well.

MohamedSabthar:
> Let's add the spec update in this PR as well.

working on it

docs/spec/spec.md
```diff
@@ -267,6 +273,189 @@ The GraphQL engine acts as the main processing unit in the Ballerina GraphQL pac
...
+    boolean validation = true;
+|};
+```
+
+#### 2.5.2 Initializing the Client
+
+The `graphql:Client` init method requires a valid URL and optional configuration to initialize the client. 
+
+```ballerina
+graphql:Client graphqlClient = check new (“http://localhost:9091/graphql”, {timeout: 10});
```
ThisaruGuruge, line 327: Let's use the port `4000` for the sample GraphQL services to maintain consistency. (Please check whether there are any other places used a port other that `4000` and update them).

docs/spec/spec.md
```diff
@@ -33,6 +33,12 @@ The conforming implementation of the specification is released and included in t
         * 2.2.5 [Service Constructor Expression](#225-service-constructor-expression)
     * 2.3 [Parser](#23-parser)
     * 2.4 [Engine](#24-engine)
+    * 2.5 [Client](#25-client)
```
ThisaruGuruge, line 36: We need to update the `Updated` field in the spec info.

docs/spec/spec.md
```diff
@@ -267,6 +273,189 @@ The GraphQL engine acts as the main processing unit in the Ballerina GraphQL pac
...
+```ballerina
+graphql:Client graphqlClient = check new (“http://localhost:9091/graphql”, {timeout: 10});
+```
+#### 2.5.3 Executing Operations
+
+The graphql client provides `execute` API to execute graphql query and mutation operations. The `execute` method of `graphql:Client` takes a GraphQL document as the required argument and sends a request to the specified backend URL seeking for a response. Further, the execute method could take the following optional arguments.
+
+- `variables` - A map containing the GraphQL variables. All the variables that may be required by the graphql document can be set via this variables argument.
+
+- `operationName` - The GraphQL operation name. If the document has more than one operation, then each operation must have a name. Single GraphQL request can only execute one operation; the operation name must be set if the document has more than one operation.
```
ThisaruGuruge, line 335: What happens when the name is not provided and there are multiple operations? Will that return an error from the client-side or will it be sent to the server? Shall we check and update the behavior here?

docs/spec/spec.md
```diff
@@ -267,6 +273,189 @@ The GraphQL engine acts as the main processing unit in the Ballerina GraphQL pac
...
+
+# Represents network-level errors.
+public type HttpError distinct (RequestError & error<record {| anydata body; |}>);
+
+# Represents GraphQL errors due to request validation.
+public type InvalidDocumentError distinct (RequestError & error<record {| ErrorDetail[]? errors; |}>);
+
+# Represents client-side data binding error.
+public type PayloadBindingError distinct (ClientError & error<record {| ErrorDetail[]? errors; |}>);
+```
```
ThisaruGuruge, line 429: Let's not just copy and paste the Ballerina code here. Instead, use point form (as sub-topics) to describe each error type.

docs/spec/spec.md
MohamedSabthar: Updated

docs/spec/spec.md
ThisaruGuruge: I think we can improve the behavior here. We can use the existing GraphQL parser to parse and validate the document before sending it out. Shall we create an issue for this?

docs/spec/spec.md
MohamedSabthar: Updated

docs/spec/spec.md
MohamedSabthar: But the syntax validation will be duplicated in both places (client and server). 
In most cases there wont be any error in the query (or once syntactically invalid query is corrected). So it will take more time to execute a valid query.
Having syntax validation in client side seems to be redundant IMO. WDYT ?

ThisaruGuruge: Point! Let's keep it as it is then.

ThisaruGuruge (approved):

PR #1015, opened 2022-10-27
Fix Socket Connection Gets Closed When the Operation Is Complete or If There Is an Error

MohamedSabthar:
## Purpose

Fixes: [#3548](https://github.com/ballerina-platform/ballerina-standard-library/issues/3548)

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [x] Updated the spec

### DIFF
--- a/ballerina-tests/tests/36_subscriptions.bal
+++ b/ballerina-tests/tests/36_subscriptions.bal
@@ -265,7 +265,7 @@ isolated function testSubscriptionWithServiceObjectsUsingSubProtocol() returns e
         websocket:ClientConfiguration config = {subProtocols: [subProtocol]};
         websocket:Client wsClient = check new(url, config);
         string messageType = subProtocol == GRAPHQL_WS ? WS_DATA : WS_NEXT;
-        check initiateConnectionInitMessage(wsClient, "1");
+        check initiateConnectionInitMessage(wsClient);
         check validateConnectionInitMessage(wsClient);
 
         check writeWebSocketTextMessage(document, wsClient, id = "1", subProtocol = subProtocol);
@@ -293,7 +293,7 @@ isolated function testSubscriptionWithVariablesUsingSubProtocol() returns error?
         websocket:Client wsClient = check new(url, config);
         string messageType = subProtocol == GRAPHQL_WS ? WS_DATA : WS_NEXT;
 
-        check initiateConnectionInitMessage(wsClient, "1");
+        check initiateConnectionInitMessage(wsClient);
         check validateConnectionInitMessage(wsClient);
 
         check writeWebSocketTextMessage(document, wsClient, variables, id = "1", subProtocol = subProtocol);
@@ -317,9 +317,10 @@ isolated function testSubscriptionWithMultipleClientsUsingSubProtocol() returns 
         websocket:Client wsClient2 = check new(url, config);
         string messageType = subProtocol == GRAPHQL_WS ? WS_DATA : WS_NEXT;
 
-        check initiateConnectionInitMessage(wsClient1, "1");
-        check initiateConnectionInitMessage(wsClient2, "2");
+        check initiateConnectionInitMessage(wsClient1);
         check validateConnectionInitMessage(wsClient1);
+
+        check initiateConnectionInitMessage(wsClient2);
         check validateConnectionInitMessage(wsClient2);
 
         check writeWebSocketTextMessage(document, wsClient1, id = "1", subProtocol = subProtocol);
@@ -354,9 +355,10 @@ isolated function testSubscriptionsWithMultipleOperationsUsingSubProtocol() retu
         websocket:Client wsClient2 = check new(url, config);
         string messageType = subProtocol == GRAPHQL_WS ? WS_DATA : WS_NEXT;
 
-        check initiateConnectionInitMessage(wsClient1, "1");
-        check initiateConnectionInitMessage(wsClient2, "2");
+        check initiateConnectionInitMessage(wsClient1);
         check validateConnectionInitMessage(wsClient1);
+
+        check initiateConnectionInitMessage(wsClient2);
         check validateConnectionInitMessage(wsClient2);
 
         check writeWebSocketTextMessage(document, wsClient1, {}, "getMessages", "1", subProtocol = subProtocol);
@@ -391,7 +393,7 @@ isolated function testSubscriptionWithFragmentsUsingSubProtocol() returns error?
         websocket:Client wsClient = check new(url, config);
         string messageType = subProtocol == GRAPHQL_WS ? WS_DATA : WS_NEXT;
 
-        check initiateConnectionInitMessage(wsClient, "1");
+        check initiateConnectionInitMessage(wsClient);
         check validateConnectionInitMessage(wsClient);
 
         check writeWebSocketTextMessage(document, wsClient, id = "1", subProtocol = subProtocol);
@@ -416,7 +418,7 @@ isolated function testInvalidSubscriptionUsingSubProtocol() returns error? {
         websocket:Client wsClient = check new(url, config);
         string messageType = subProtocol == GRAPHQL_WS ? WS_DATA : WS_ERROR;
         json responsePayload = check getJsonContentFromFile("subscription_invalid_field.json");
-        check initiateConnectionInitMessage(wsClient, "1");
+        check initiateConnectionInitMessage(wsClient);
         check validateConnectionInitMessage(wsClient);
 
         json payload = {query: document};
@@ -438,7 +440,7 @@ isolated function testSubscriptionFunctionWithErrorsUsingSubProtocol() returns e
         websocket:Client wsClient = check new(url, config);
         string messageType = subProtocol == GRAPHQL_WS ? WS_DATA : WS_ERROR;
 
-        check initiateConnectionInitMessage(wsClient, "1");
+        check initiateConnectionInitMessage(wsClient);
         check validateConnectionInitMessage(wsClient);
 
         check writeWebSocketTextMessage(document, wsClient, id = "1", subProtocol = subProtocol);
@@ -458,7 +460,7 @@ isolated function testConnectionInitMessage() returns error? {
     foreach string subProtocol in subProtocols {
         websocket:ClientConfiguration config = {subProtocols: [subProtocol]};
         websocket:Client wsClient = check new(url, config);
-        check initiateConnectionInitMessage(wsClient, "1");
+        check initiateConnectionInitMessage(wsClient);
         check validateConnectionInitMessage(wsClient);
     }
 }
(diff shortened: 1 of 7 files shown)


ballerina-tests/tests/36_subscriptions.bal
```diff
@@ -601,3 +585,33 @@ isolated function testErrorsInStreams() returns error? {
     expectedPayload = {data: {evenNumber: 6}};
     check validateWebSocketResponse(wsClient, expectedPayload);
 }
+
+@test:Config {
+    groups: ["sub_protocols", "subscriptions", "a"]
```
DimuthuMadushan, line 590:
```suggestion
    groups: ["sub_protocols", "subscriptions"]
```
```suggestion
    groups: ["sub_protocols", "subscriptions", "a"]
```

ballerina/websocket_service.bal
```diff
@@ -24,17 +24,17 @@ isolated service class WsService {
...
     private final Context context;
-    private string[] activeConnections;
+    private map<string> activeConnections;
     private final readonly & map<string> customHeaders;
     private boolean initiatedConnection;
 
-    isolated function init(Engine engine, __Schema & readonly schema,  map<string> & readonly customHeaders,
-                           Context context) {
+    isolated function init(Engine engine, __Schema & readonly schema, map<string> & readonly customHeaders,
+            Context context) {
```
DimuthuMadushan, line 32: Incorrect indentation?

ballerina/websocket_service.bal
```diff
@@ -58,7 +58,7 @@ isolated service class WsService {
         }
 
         WSPayload|json|error wsPayload = self.customHeaders != {}
-                                        ? wsText.cloneWithType(WSPayload) : value:fromJsonString(text);
+            ? wsText.cloneWithType(WSPayload) : value:fromJsonString(text);
```
DimuthuMadushan, line 61: Any reason to change the indentation here?

changelog.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Fixed
 - [[#3556] Fix Returning Table from a Resolver Resulting in Runtime Error](https://github.com/ballerina-platform/ballerina-standard-library/issues/3556)
 - [[#3565] Fix Service Incorrectly Closing Connection when a Stream Returns an Error](https://github.com/ballerina-platform/ballerina-standard-library/issues/3565)
+- [[#3548] Fix Socket Connection Gets Closed When the Operation Is Complete or If There Is an Error](https://github.com/ballerina-platform/ballerina-standard-library/issues/3548)
```
DimuthuMadushan, line 12:
```suggestion
- [[#3548] Fix Socket Connection Gets Closed when the Operation is Complete or If There is an Error](https://github.com/ballerina-platform/ballerina-standard-library/issues/3548)
```

ballerina/websocket_service.bal
MohamedSabthar: No, this is formatted with ballerina

ballerina/websocket_service.bal
MohamedSabthar: Since this file has lot of changes I have formatted it with ballerina. The previous one is not properly indented

ballerina/websocket_service.bal
DimuthuMadushan: This is not what we used in other places. we have used the previous indentation in the function definition.

changelog.md
MohamedSabthar: Updated

ballerina-tests/tests/36_subscriptions.bal
MohamedSabthar: Removed

ballerina/websocket_service.bal
```diff
@@ -24,17 +24,17 @@ isolated service class WsService {
...
-    isolated function init(Engine engine, __Schema & readonly schema,  map<string> & readonly customHeaders,
-                           Context context) {
+    isolated function init(Engine engine, __Schema & readonly schema, map<string> & readonly customHeaders,
+            Context context) {
         self.engine = engine;
         self.schema = schema;
         self.context = context;
         self.customHeaders = customHeaders;
-        self.activeConnections = [];
+        self.activeConnections = {};
```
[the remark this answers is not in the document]
MohamedSabthar: Timeout handled here: https://github.com/ballerina-platform/module-ballerina-graphql/blob/364b390f4155c368c6c023347e76c0e1ac8b891c/ballerina/websocket_service.bal#L44

ballerina/websocket_service.bal
```diff
@@ -24,17 +24,17 @@ isolated service class WsService {
     private final Engine engine;
     private final readonly & __Schema schema;
     private final Context context;
-    private string[] activeConnections;
+    private map<string> activeConnections;
```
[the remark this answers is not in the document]
MohamedSabthar: Search operation is much faster with map compared to an array

ballerina/websocket_service.bal
```diff
@@ -87,28 +87,25 @@ isolated service class WsService {
                             closeConnection(caller, 4401, "Unauthorized");
                             return;
                         }
-                        if self.activeConnections.indexOf(connectionId) !is () {
+                        if self.activeConnections.hasKey(connectionId) {
```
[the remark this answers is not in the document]
MohamedSabthar: According to the spec, we should close the connection if there is 2 subscription operation with same id
```If there is already an active subscriber for an operation matching the provided ID, regardless of the operation type, the server **must** close the socket immediately with the event 4409: Subscriber for <unique-operation-id> already exists```

File: ballerina/websocket_service.bal (hunk 6 above)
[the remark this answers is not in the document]
Comment (ThisaruGuruge) [reply ballerina/websocket_service.bal:27]: But this increases memory usage, right?

File: ballerina/websocket_service.bal (hunk 7 above)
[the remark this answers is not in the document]
Comment (ThisaruGuruge) [reply ballerina/websocket_service.bal:90]: That's not what I asked. We are checking each request for a connection ID, but it is sent during the `SUBSCRIBE` operation. So, there's no need to check the connection id before every operation.

File: ballerina/websocket_service.bal (hunk 2 above)
Comment (ThisaruGuruge) [reply ballerina/websocket_service.bal:32]: Ballerina formatter does weird formatting when a single line breaks into two. Let's stick with the existing indentations for now.

File: ballerina/websocket_service.bal (hunk 6 above)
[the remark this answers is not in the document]
Comment (MohamedSabthar) [reply ballerina/websocket_service.bal:27]: That's true same string will be stored as key and value, in that case we could use `map<()>`

File: ballerina/websocket_service.bal (hunk 7 above)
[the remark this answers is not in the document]
Comment (MohamedSabthar) [reply ballerina/websocket_service.bal:90]: Now I get the point it's related to the logic in line 66. I'll update the PR

File: ballerina/websocket_service.bal (hunk 6 above)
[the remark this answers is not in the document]
Comment (ThisaruGuruge) [reply ballerina/websocket_service.bal:27]: +1 for using `map<()>`

File: ballerina/websocket_service.bal (hunk 7 above)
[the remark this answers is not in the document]
Comment (MohamedSabthar) [reply ballerina/websocket_service.bal]: updated

File: ballerina/websocket_service.bal (hunk 2 above)
Comment (MohamedSabthar) [reply ballerina/websocket_service.bal]: reverted

File: ballerina/websocket_service.bal [hunk 8]
```diff
@@ -57,19 +57,17 @@ isolated service class WsService {
...
+            ? wsText.cloneWithType(WSPayload) : value:fromJsonString(text);
+        if payload is error {
+            return self.handleError(caller, payload);
         }
-        string wsType = wsPayload is WSPayload ? <string>wsPayload.'type : DEFAULT_VALUE;
-        string connectionId = wsPayload is WSPayload && wsPayload?.id !is () ? <string>wsPayload?.id : DEFAULT_VALUE;
-
         if !self.customHeaders.hasKey(WS_SUB_PROTOCOL) {
-            return self.handleSubscriptionRequest(caller, connectionId, wsPayload);
+            return self.handleSubscriptionRequest(caller, DEFAULT_VALUE, payload);
```
ThisaruGuruge, line 66: In this case, do we need the `connectionId`?

MohamedSabthar: No we could change the handleSubscriptionRequest's id parameter to have `string?` type

ThisaruGuruge: Yeah, that'd be better.

ballerina/websocket_service.bal
```diff
@@ -81,34 +79,37 @@ isolated service class WsService {
                 }
             }
             WS_SUBSCRIBE|WS_START => {
+                if wsPayload.id is () {
+                    return self.handleIdNotPresentInPayload(caller);
+                }
+                string connectionId = <string>wsPayload.id;
```
ThisaruGuruge, line 85: We should avoid castings like this. It is considered bad practice in Ballerina. The better way to handle this is this:

```suggestion
                string? connectionId = wsPayload.id;
                if connectionId is () {
                    return self.handleIdNotPresentInPayload(caller);
                }
                // ...
```

Assigning `wsPayload.id` to a local variable will nil-lift after returning.

ballerina/websocket_service.bal
```diff
@@ -57,19 +57,17 @@ isolated service class WsService {
...
-        string wsType = wsPayload is WSPayload ? <string>wsPayload.'type : DEFAULT_VALUE;
-        string connectionId = wsPayload is WSPayload && wsPayload?.id !is () ? <string>wsPayload?.id : DEFAULT_VALUE;
-
         if !self.customHeaders.hasKey(WS_SUB_PROTOCOL) {
-            return self.handleSubscriptionRequest(caller, connectionId, wsPayload);
+            return self.handleSubscriptionRequest(caller, DEFAULT_VALUE, payload);
         }
 
-        match wsType {
+        WSPayload wsPayload = <WSPayload>payload;
```
ThisaruGuruge, line 69: Why do we need this casting here?

MohamedSabthar: In line 65 we have check for `!self.customHeaders.hasKey(WS_SUB_PROTOCOL)` instead of checking `payload is json` this makes the resulting type to have `WSPayload|json`. Casting is done to eliminate json type from the union.

We also could remove the casting by checking `payload is json` in line 65. but this could make code slightly unclear

ballerina/websocket_service.bal
MohamedSabthar: updated

ballerina/websocket_service.bal
ThisaruGuruge: We can type-lift the json type as well then.

MohamedSabthar: WSPayload|json is ambiguous type type. 

the other way we could avoid casting is to replace the casting with following if block
```
if wsPayload !is WSPayload {
   return;
}
```
will update the PR with this

MohamedSabthar: updated

ThisaruGuruge (approved):

PR #1024, opened 2022-11-01
Fix Resolver returning null when a record has a service object as its field 

MohamedSabthar:
## Purpose

Fixes: [#3466](https://github.com/ballerina-platform/ballerina-standard-library/issues/3466)
Fixes: [#3579](https://github.com/ballerina-platform/ballerina-standard-library/issues/3579)

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [ ] ~~Updated the spec~~


ballerina-tests/tests/07_service_objects.bal
```diff
@@ -77,3 +77,106 @@ isolated function testComplexService() returns error? {
...
+                },
+                {
+                    product: {
+                        id: "3"
+                    },
+                    score: 20
+                }
+            ]
+        }
+    };
```
ThisaruGuruge, line 133: Shall we move these expected values to the resources directory?

ballerina/response_generator.bal
```diff
@@ -43,16 +43,16 @@ class ResponseGenerator {
...
-        if parentValue is map<anydata> {
-            if isMap(parentValue) {
-                return self.getResultFromMap(<map<anydata>>parentValue, parentNode);
-            }
+        if parentValue is map<anydata> && isMap(parentValue) {
+            return self.getResultFromMap(parentValue, parentNode);
+        }
+        if parentValue is map<any> {
             return self.getResultFromRecord(parentValue, parentNode);
         }
```
ThisaruGuruge, line 51: This condition does not look right. 

The `is map<anydata>` returns true for both `map` and `record` types. That's why we need an additional check `isMap()` using native code. Now, in this case, we have 4 cases. 

- `map` with `anydata`
- `map` with `any`
- `record` with `anydata`
- `record` with `any`

But I think we can treat `anydata` and `any` similarly here. So, what we really need is 

```suggestion
        if parentValue is map<any> {
            if isMap(parentValue) {
                return self.getResultFromMap(parentValue, parentNode);
            }
            return self.getResultFromRecord(parentValue, parentNode);
        }
```

ballerina-tests/tests/records.bal
```diff
@@ -243,3 +243,9 @@ public type Languages record {|
 public type CovidEntry record {|
     string isoCode;
 |};
+
+public type Review record {
+  Product product;
+  int score;
+  string description;
+};
```
DimuthuMadushan, line 251: Shall we use closed record here if possible?

ballerina-tests/tests/07_service_objects.bal
MohamedSabthar: Updated

ballerina-tests/tests/records.bal
MohamedSabthar: Updated

ballerina/response_generator.bal
MohamedSabthar: Fixed with [93433b7](https://github.com/ballerina-platform/module-ballerina-graphql/pull/1024/commits/93433b74c3f72d4ab79e469bae9bf9a73c58114b)

ballerina-tests/Dependencies.toml
```diff
@@ -105,7 +105,8 @@ dependencies = [
 	{org = "ballerina", name = "websocket"}
 ]
 modules = [
-	{org = "ballerina", packageName = "graphql_tests", moduleName = "graphql_tests"}
+	{org = "ballerina", packageName = "graphql_tests", moduleName = "graphql_tests"},
+	{org = "ballerina", packageName = "graphql_tests", moduleName = "graphql_tests.graphql"}
```
ThisaruGuruge, line 109: Why is there a new submodule in the graphql_tests package?

MohamedSabthar:
> The second issue mentioned in the PR is not correctly linked to the issue. When we are sending fixes to multiple issues, we should use the following notation:
> 
> ```
> Fixes: <Issue #1>
> Fixes: <Issue #2>
> ...
> ```
> 
> Also, make sure to add the correct resolution label for an issue, when sending a fix.

Noted.

ballerina-tests/Dependencies.toml
MohamedSabthar: Seems there a graphql folder crated inside the graphql_tests package. Removed now

ThisaruGuruge (approved):

PR #1152, opened 2023-01-05
Add Block String Documentation Generation in SDL Schema Generator

DimuthuMadushan:
## Purpose

Fixes: [#3819](https://github.com/ballerina-platform/ballerina-standard-library/issues/3819)

## Examples
N/A

## Checklist
- [x] Linked to an issue
- [ ] ~Updated the changelog~
- [ ] ~Added tests~
- [ ] ~Updated the spec~

### DIFF
--- a/commons/src/main/java/io/ballerina/stdlib/graphql/commons/utils/SdlSchemaStringGenerator.java
+++ b/commons/src/main/java/io/ballerina/stdlib/graphql/commons/utils/SdlSchemaStringGenerator.java
@@ -71,7 +71,7 @@ public class SdlSchemaStringGenerator {
     private static final String DOUBLE_INDENTATION = INDENTATION + INDENTATION;
     private static final String EMPTY_STRING = "";
     private static final String COMMA_SIGN = ",";
-    private static final String AND_SIGN = "&";
+    private static final String AMPERSAND_SIGN = "&";
     private static final String PIPE_SIGN = "|";
     private static final String SPACE = " ";
 
@@ -172,7 +172,7 @@ public class SdlSchemaStringGenerator {
         if (description == null) {
             return EMPTY_STRING;
         } else {
-            String[] lines = removeLeadingNewLines(description).split(LINE_SEPARATOR);
+            String[] lines = description.trim().split(LINE_SEPARATOR);
             if (lines.length == 1) {
                 return getFormattedString(DESC_FORMAT, getFormattedString(DOCUMENT_FORMAT, EMPTY_STRING, lines[0]));
             } else {
@@ -268,14 +268,14 @@ public class SdlSchemaStringGenerator {
         for (Type interfaceType : type.getInterfaces()) {
             interfaces.add(interfaceType.getName());
         }
-        return getFormattedString(IMPLEMENT_FORMAT, String.join(SPACE + AND_SIGN + SPACE, interfaces));
+        return getFormattedString(IMPLEMENT_FORMAT, String.join(SPACE + AMPERSAND_SIGN + SPACE, interfaces));
     }
 
     private static String createFieldDescription(String description) {
         if (description == null) {
             return EMPTY_STRING;
         } else {
-            String[] lines = removeLeadingNewLines(description).split(LINE_SEPARATOR);
+            String[] lines = description.trim().split(LINE_SEPARATOR);
             if (lines.length == 1) {
                 return getFormattedString(DESC_FORMAT, getFormattedString(DOCUMENT_FORMAT, INDENTATION, lines[0]));
             } else {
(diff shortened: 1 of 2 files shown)


commons/src/main/java/io/ballerina/stdlib/graphql/commons/utils/SdlSchemaStringGenerator.java
```diff
@@ -67,10 +68,12 @@ public class SdlSchemaStringGenerator {
     //Schema delimiters
     private static final String LINE_SEPARATOR = System.getProperty("line.separator");
     private static final String INDENTATION = "  ";
+    private static final String DOUBLE_INDENTATION = INDENTATION + INDENTATION;
     private static final String EMPTY_STRING = "";
-    private static final String COMMA_SIGN = ", ";
-    private static final String AND_SIGN = " & ";
+    private static final String COMMA_SIGN = ",";
+    private static final String AND_SIGN = "&";
```
ThisaruGuruge, line 74: `AMPERSAND_SIGN`

commons/src/main/java/io/ballerina/stdlib/graphql/commons/utils/SdlSchemaStringGenerator.java
```diff
@@ -331,4 +375,12 @@ private static Boolean isDefaultDirective(Directive directive) {
...
         return String.format(format, (Object[]) args);
     }
+
+    private static String removeLeadingNewLines(String description) {
+        int i = 0;
+        while (i < description.length() && String.valueOf(description.charAt(i)).equals(LINE_SEPARATOR)) {
+            i++;
+        }
+        return description.substring(i);
+    }
```
ThisaruGuruge, line 385: `description.trim()`?

commons/src/main/java/io/ballerina/stdlib/graphql/commons/utils/SdlSchemaStringGenerator.java
```diff
@@ -216,10 +223,23 @@ private static String createArgs(List<InputValue> inputValues) {
...
+            return getFormattedString(ARGS_FORMAT, LINE_SEPARATOR, String.join(LINE_SEPARATOR, args),
+                    LINE_SEPARATOR, INDENTATION);
+        } else {
+            for (InputValue arg : inputValues) {
+                args.add(getFormattedString(ARGS_VALUE_FORMAT, EMPTY_STRING, EMPTY_STRING, arg.getName(),
+                        createArgType(arg)));
+            }
+            return getFormattedString(ARGS_FORMAT, EMPTY_STRING, String.join(COMMA_SIGN + SPACE, args), EMPTY_STRING,
+                    EMPTY_STRING);
         }
```
ThisaruGuruge, line 242: Shouldn't this be handled in the `createArgDescription` method? Here, the `inputValues` array is iterated twice, once to check `hasDescription`, another time to get the description. We can avoid this by handling it inside the description generation.

commons/src/main/java/io/ballerina/stdlib/graphql/commons/utils/SdlSchemaStringGenerator.java
```diff
@@ -331,4 +375,12 @@ private static Boolean isDefaultDirective(Directive directive) {
     private static String getFormattedString(String format, String... args) {
         return String.format(format, (Object[]) args);
     }
+
+    private static String removeLeadingNewLines(String description) {
+        int i = 0;
+        while (i < description.length() && String.valueOf(description.charAt(i)).equals(LINE_SEPARATOR)) {
```
MohamedSabthar, line 381: When LINE_SEPARATOR takes the value "\r\n" will this work ? because we are using `charAt(i)` here which check only on character at a time

commons/src/main/java/io/ballerina/stdlib/graphql/commons/utils/SdlSchemaStringGenerator.java
DimuthuMadushan: If there is a single argument description, the entire argument block format and each single argument format will be changed. Therefore used two iterations here. If this is moved into `createDescription`, it will be complex to handle this scenario.

commons/src/main/java/io/ballerina/stdlib/graphql/commons/utils/SdlSchemaStringGenerator.java
DimuthuMadushan: Removed this function

commons/src/main/java/io/ballerina/stdlib/graphql/commons/utils/SdlSchemaStringGenerator.java
DimuthuMadushan: Replaced this with `trim()`

commons/src/main/java/io/ballerina/stdlib/graphql/commons/utils/SdlSchemaStringGenerator.java
ThisaruGuruge: Can you include an example of that?

DimuthuMadushan: 1. if there is no arg descriptions
`name(id: String!, age: Int!, isAdult: Boolean!): String!`

2. if there is at least one arg description
```
name(
    id: String!
    """
    This should be an integer
    This is a Non-Null type argument
    """
    age: Int!
    isAdult: Boolean!
  ): String!
```

ThisaruGuruge (approved):

PR #1190, opened 2023-01-30
Fix testAlreadyExistingSubscriber failing intermittently

MohamedSabthar:
## Purpose

Fixes: https://github.com/ballerina-platform/ballerina-standard-library/issues/4009

## Checklist
- [x] Linked to an issue
- [ ] ~~Updated the changelog~~
- [ ] ~~Added tests~~
- [ ] ~~Updated the spec~~

### DIFF
--- a/ballerina-tests/tests/36_subscriptions.bal
+++ b/ballerina-tests/tests/36_subscriptions.bal
@@ -355,8 +355,19 @@ function testAlreadyExistingSubscriber() returns error? {
     check sendSubscriptionMessage(wsClient, document, clientId);
     check sendSubscriptionMessage(wsClient, document, clientId);
 
+    json|error response;
+    while true {
+        response = readMessageExcludingPingMessages(wsClient);
+        if response is error {
+            break;
+        }
+        json|error id = response.id;
+        if id is error {
+            test:assertFail(string `Expected json with id found: ${response.toString()}`);
+        }
+    }
     string expectedErrorMsg = "Subscriber for " + clientId + " already exists: Status code: 4409";
-    validateConnectionClousureWithError(wsClient, expectedErrorMsg);
+    test:assertEquals((<error>response).message(), expectedErrorMsg);
 }
 
 @test:Config {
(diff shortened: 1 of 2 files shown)


ballerina-tests/tests/utils.bal
```diff
@@ -168,6 +168,9 @@ isolated function validateCompleteMessage(websocket:Client wsClient, string id =
 
 isolated function validateConnectionClousureWithError(websocket:Client wsClient, string expectedErrorMsg) {
     json|error response = readMessageExcludingPingMessages(wsClient);
-    test:assertTrue(response is error);
-    test:assertEquals((<error>response).message(), expectedErrorMsg);
+    if response is error {
+        test:assertEquals((<error>response).message(), expectedErrorMsg);
```
ThisaruGuruge, line 172: No need of casting here.
```suggestion
        test:assertEquals(response.message(), expectedErrorMsg);
```

ballerina-tests/tests/36_subscriptions.bal
```diff
@@ -355,8 +355,19 @@ function testAlreadyExistingSubscriber() returns error? {
...
+    json|error response;
+    while true {
+        response = readMessageExcludingPingMessages(wsClient);
+        if response is error {
+            break;
+        }
+        json|error id = response.id;
+        if id is error {
+            test:assertFail(string `Expected json with id found: ${response.toString()}`);
+        }
```
ThisaruGuruge, line 367: Can this lead into an infinite loop? What if we don't get an error?

MohamedSabthar: Since server side uses multiplexing it will return an error for the second subscription operation

ballerina-tests/tests/utils.bal
MohamedSabthar: As I checked type narrowing doesn't work here

MohamedSabthar: Updated

ballerina-tests/tests/36_subscriptions.bal
MohamedSabthar: Updated fail after 4 iteration

ballerina-tests/tests/utils.bal
ThisaruGuruge: Seems like. this isn't updated properly.

MohamedSabthar: changes pushed

ThisaruGuruge (approved):

PR #1195, opened 2023-01-31
Fix GraphQL Resolvers not Passing Enum Values Correctly when Used as Input Parameters

ThisaruGuruge:
## Purpose

$Subject

Fixes: [#3721](https://github.com/ballerina-platform/ballerina-standard-library/issues/3721)

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [ ] ~Updated the spec~


ballerina-tests/tests/test_services.bal
```diff
@@ -742,6 +742,34 @@ service /special_types on specialTypesTestListener {
     resource function get company() returns Company {
         return company;
     }
+
+    resource function get month(Month month) returns string {
```
MohamedSabthar, line 746: Isn't it necessary to return the Month in order to verify the bug fix?

ThisaruGuruge: Yes, otherwise the passed string would be `JANUARY`, `February`, etc.

DimuthuMadushan: If we return the `Month` type here(just like the example in the issue), we can verify the output value as well.

ThisaruGuruge: Right. I think you are correct. Will update the code.

ThisaruGuruge: Fixed with 008f37bb942e035068628ce8a344c0ff201ddfe7

DimuthuMadushan (approved):

MohamedSabthar (approved):

PR #1215, opened 2023-02-09
Fix context and Field Object Parameter Ordering Issue

ThisaruGuruge:
## Purpose

This is to fix the issue where the `graphql:Context` and `graphql:Field` object parameters are not populated properly when they are added after the other parameters. Found this issue when testing with the SwanLake Update 4 RC1. 

An issue was not created because the issue is part of the feature that is intended to ship with this release. This issue hasn't been affected anything. 

## Checklist
- [ ] ~Linked to an issue~
- [ ] ~Updated the changelog~
- [x] Added tests
- [ ] ~Updated the spec~

### DIFF
--- a/ballerina-tests/tests/resources/documents/field_object_parameter_order.graphql
+++ b/ballerina-tests/tests/resources/documents/field_object_parameter_order.graphql
@@ -1,0 +1,12 @@
+query FieldObjectParameterOrder1 {
+    student(name: "Skinny Pete", id: 100) {
+        name
+    }
+}
+
+query FieldObjectParameterOrder2 {
+    student(name: "Skinny Pete", id: 100) {
+        name
+        id
+    }
+}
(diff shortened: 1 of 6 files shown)


ballerina-tests/tests/resources/documents/field_object_parameter_order.graphql
```diff
@@ -0,0 +1,12 @@
+query FieldObjectParameterOrder {
```
DimuthuMadushan, line 1:
```suggestion
query FieldObjectParameterOrder1 {
```

DimuthuMadushan (approved):

ballerina-tests/tests/resources/documents/field_object_parameter_order.graphql
ThisaruGuruge: Fixed

PR #1297, opened 2023-03-16
Improve GraphQL Validation Logics

DimuthuMadushan:
## Purpose

Fixes: [#4206](https://github.com/ballerina-platform/ballerina-standard-library/issues/4206)

## Examples
N/A

## Checklist
- [x] Linked to an issue
- [ ] ~Updated the changelog~
- [x] Added tests
- [ ] ~Updated the spec~

### DIFF
--- a/ballerina/engine.bal
+++ b/ballerina/engine.bal
@@ -53,11 +53,8 @@ isolated class Engine {
             return result;
         }
 
-        OutputObject|parser:DocumentNode validationResult = self.validateDocument(result, variables);
-        if validationResult is OutputObject {
-            return validationResult;
-        }
-        return self.getOperation(validationResult, operationName);
+        OutputObject|parser:OperationNode validationResult = self.validateDocument(result, operationName, variables);
+        return validationResult;
     }
 
     isolated function getResult(parser:OperationNode operationNode, Context context, any|error result = ())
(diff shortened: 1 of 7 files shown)


ballerina/engine.bal
```diff
@@ -97,22 +94,28 @@ isolated class Engine {
...
+    returns OutputObject|parser:OperationNode {
+        ErrorDetail[]|NodeModifierContext validationResult =
+            self.parallellyValidateDocument(parseResult, operationName, variables);
         if validationResult is ErrorDetail[] {
             return getOutputObjectFromErrorDetail(validationResult);
         } else {
             DocumentNodeModifierVisitor documentNodeModifierVisitor = new (validationResult);
             parseResult.document.accept(documentNodeModifierVisitor);
-            return documentNodeModifierVisitor.getDocumentNode();
+            return documentNodeModifierVisitor.getDocumentNode().getOperations()[0];
```
ThisaruGuruge, line 106: I think returning the `OperationNode` from the `validateDocument` method is kinda incorrect. Shall return the `DocumentNode` here and access the operation node in the `validate` method? When accessing, better to put a comment about why we directly access the 0th operation node, because, at the first glance, it is not easily understandable why we use the 0th operation every time. Took some time for me here as well.

DimuthuMadushan: Fixed in https://github.com/ballerina-platform/module-ballerina-graphql/pull/1297/commits/2f3f897d6bbbabd7457d6cfe38dead3f347cd818

ThisaruGuruge (approved):

PR #1349, opened 2023-04-03
Introduce GraphQL Interceptor Configs

DimuthuMadushan:
## Purpose

Fixes: [#4254](https://github.com/ballerina-platform/ballerina-standard-library/issues/4254)

## Examples
```ballerina
@graphql:InterceptorConfig {
    global: true
}
readonly service class ServiceInterceptor {
    isolated remote function execute() returns anydata|error{
        //...
    }
}
```

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [x] Updated the spec


docs/spec/spec.md
```diff
@@ -2001,6 +2003,31 @@ service on new graphql:Listener(9090) {
 }
 ```
 
+### 10.3 Interceptor Configuration
+
+The configurations stated in the `graphql:InterceptorConfig`, are used to change the behavior of a particular GraphQL interceptor.
+
+#### 10.3.1 Scope Configuration
+
+The field `global` is used to configure the scope of the interceptor. If the `global` field is set as `true`, the interceptor will be applied to each field and subfield. If the flag is set as `false`, the interceptor will be applied only to the particular type. By default, the `global` flag is set as `true`.
```
ThisaruGuruge, line 2012:
```suggestion
The field `global` is used to configure the scope of the interceptor. If the `global` field is set as `true`, the interceptor will be applied to each field and subfield of the service. If the flag is set as `false`, the interceptor will be applied only to the fields of the type, but not to the subfields of the type. By default, the `global` flag is set as `true`.
```

ballerina/context.bal
```diff
@@ -138,25 +138,27 @@ public isolated class Context {
...
-                if 'field.getFieldInterceptors().length() > nextFieldInterceptor {
-                    readonly & Interceptor next = 'field.getFieldInterceptors()[nextFieldInterceptor];
-                    self.nextInterceptor += 1;
-                    return next;
+        Engine? engine = self.getEngine();
+        if engine is Engine {
+            (readonly & Interceptor)[] interceptors = engine.getInterceptors();
+            if interceptors.length() > self.getInterceptorCount() {
+                (readonly & Interceptor) next = interceptors[self.getInterceptorCount()];
+                if !isGlobalInterceptor(next) && 'field.getPath().length() > 1 {
```
ThisaruGuruge, line 146: Is this logic accurate? We need a better way to figure out how to omit the non-global interceptors without considering the path. 

We might need to apply this to other service types as well. This logic won't work in those scenarios.

I think we might have to keep track of the interceptors and their respective types somewhere else. If this is too complex to handle, we can go ahead with this logic for now and improve later on.

ThisaruGuruge: +1

ThisaruGuruge (approved):

PR #1440, opened 2023-05-29
Improve the Return Type Validation in GraphQL Interceptors

DimuthuMadushan:
## Purpose

Fixes: [#4255](https://github.com/ballerina-platform/ballerina-standard-library/issues/4255)
Fixes: [#4490](https://github.com/ballerina-platform/ballerina-standard-library/issues/4490)

## Examples
N/A

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [ ] ~Updated the spec~

### DIFF
--- a/ballerina/response_formatter.bal
+++ b/ballerina/response_formatter.bal
@@ -188,10 +188,8 @@ class ResponseFormatter {
     isolated function getFieldValue(Data data, parser:FieldNode fieldNode) returns anydata|anydata[] {
         if data.hasKey(fieldNode.getAlias()) {
             return data[fieldNode.getAlias()];
-        } else if data.hasKey(fieldNode.getName()) {
-            return data[fieldNode.getName()];
         }
-        return;
+        return data[fieldNode.getName()];
     }
 }
 
--- a/ballerina/response_formatter.bal
+++ b/ballerina/response_formatter.bal
@@ -64,7 +64,7 @@ class ResponseFormatter {
                 } else if selection is parser:FieldNode {
                     __Type fieldType = self.getFieldType(selection.getName(), parentType, onType);
                     anydata|anydata[] fieldResult = ();
-                    if data.hasKey(selection.getAlias()) {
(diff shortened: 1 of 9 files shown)


ballerina/response_formatter.bal
```diff
@@ -64,7 +64,7 @@ class ResponseFormatter {
                 } else if selection is parser:FieldNode {
                     __Type fieldType = self.getFieldType(selection.getName(), parentType, onType);
                     anydata|anydata[] fieldResult = ();
-                    if data.hasKey(selection.getAlias()) {
+                    if data.hasKey(selection.getAlias()) || data.hasKey(selection.getName()) {
```
ThisaruGuruge, line 67: Isn't this lead to a bug? What if a certain field is required twice, once with an alias and once without an alias? Do we have a test for that?

ballerina/response_formatter.bal
```diff
@@ -178,6 +184,15 @@ class ResponseFormatter {
             return <__Field>getFieldFromFieldArray(fields, fieldName);
         }
     }
+
+    isolated function getFieldValue(Data data, parser:FieldNode fieldNode) returns anydata|anydata[] {
+        if data.hasKey(fieldNode.getAlias()) {
+            return data[fieldNode.getAlias()];
+        } else if data.hasKey(fieldNode.getName()) {
```
ThisaruGuruge, line 191: Do we need the `else if` block here? If `data.hasKey(fieldNode.getAlias())` fails, it must have `fieldNode.getName()` key right?

ballerina/response_formatter.bal
DimuthuMadushan: Do you mean a query like this? https://github.com/ballerina-platform/module-ballerina-graphql/pull/1440/files#diff-0219b63c80c6cac590137488c0e68e6491fe15e99f0305e41dd5643090116d17

ThisaruGuruge: That's exactly what I was looking for!

ThisaruGuruge (approved):
LGTM

PR #1560, opened 2023-07-27
Make Gradle to do not track incremental builds 

gabilang:
## Purpose
$title
Fixes:

## Checklist
- [ ] Linked to an issue
- [ ] Updated the changelog
- [ ] Added tests
- [ ] Updated the spec
- [ ] Checked native-image compatibility
- [ ] No `commons` package changes (if there are any, please update the GraphQL version in [GraphQL tools](https://github.com/ballerina-platform/graphql-tools))

### DIFF
--- a/ballerina/build.gradle
+++ b/ballerina/build.gradle
@@ -96,6 +96,7 @@ publishing {
 
 updateTomlFiles.dependsOn copyStdlibs
 
+build.doNotTrackState("build needs to re-run every time")
 build.dependsOn "generatePomFileForMavenPublication"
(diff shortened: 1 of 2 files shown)


ballerina/build.gradle
```diff
@@ -96,6 +96,7 @@ publishing {
 
 updateTomlFiles.dependsOn copyStdlibs
 
+build.doNotTrackState("build needs to re-run every time")
```
ThisaruGuruge, line 99: Why are we re-running the build every time? One of the main reasons to use Gradle was the build caching, which improves the build time and makes the life easy for the developers.

PR #1579, opened 2023-08-15
Add compile-time subgraph schema generation

MohamedSabthar:
## Purpose

Fixes: https://github.com/ballerina-platform/ballerina-standard-library/issues/4337

## Checklist
- [x] Linked to an issue
- [ ] Updated the changelog
- [x] Added tests
- [x] Updated the spec
- [ ] No `commons` package changes (if there are any, please update the GraphQL version in [GraphQL tools](https://github.com/ballerina-platform/graphql-tools))

### DIFF
--- a/docs/spec/spec.md
+++ b/docs/spec/spec.md
@@ -966,7 +966,7 @@ In Ballerina, `distinct` `service` objects can be used to define GraphQL interfa
 
 Non-distinct `service` objects and `service` classes can not be used to define or implement GraphQL interfaces.
 
->**Note**: In order to be recognized as GraphQL objects or interfaces, the Ballerina `service` `object`s and `service` `class`es should be defined within the same module as the GraphQL service.
+>**Note**: In order to be recognized as GraphQL objects or interfaces, the Ballerina `service` `object`s and `service` `class`es must be defined within the same module as the GraphQL service.
 
 ###### Example: Interfaces
 ```ballerina


compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/diagnostics/DiagnosticMessage.java
```diff
@@ -96,9 +96,17 @@ public enum DiagnosticMessage {
...
     WARNING_201("invalid usage of @deprecated directive found in ''{0}''. Input object field(s) deprecation "
-                    + "is not supported by the current GraphQL spec."),
+                        + "is not supported by the current GraphQL spec."),
     WARNING_202("unable to validate ''{0}'' configuration of the GraphQL field ''{1}''. Pass a string literal to "
-                    + "the ''{0}'' configuration to resolve this warning");
+                        + "the ''{0}'' configuration to resolve this warning"),
+    WARNING_203("unable to validate the @subgraph:Entity annotation. This could lead to an invalid generated subgraph"
+                        + " schema. Pass key-value pairs in the @subgraph:Entity annotation to resolve this warning."),
+    WARNING_204("unable to validate the @subgraph:Entity annotation. This could lead to an invalid generated subgraph"
+                        + " schema. Provide an array of string literal value for the ''{0}'' field in the"
```
ThisaruGuruge, line 105:
```suggestion
                        + " schema. Provide an array of string literal values for the ''{0}'' field in the"
```

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/service/validator/ServiceValidator.java
```diff
@@ -132,6 +138,75 @@ public void validate() {
...
+                addDiagnostic(CompilationDiagnostic.PROVIDE_KEY_VALUE_PAIR_FOR_ENTITY_ANNOTATION, fieldNode.location());
+                continue;
+            }
+            SpecificFieldNode specificFieldNode = (SpecificFieldNode) fieldNode;
+            Node fieldNameNode = specificFieldNode.fieldName();
+            if (fieldNameNode.kind() != SyntaxKind.IDENTIFIER_TOKEN) {
+                continue;
+            }
+            IdentifierToken fieldNameToken = (IdentifierToken) fieldNameNode;
+            String fieldName = fieldNameToken.text().trim();
```
ThisaruGuruge, line 182: Do we need the `trim()` here?

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/service/validator/ServiceValidator.java
```diff
@@ -132,6 +138,75 @@ public void validate() {
...
+            if (fieldName.equals(KEY)) {
```
ThisaruGuruge, line 183:
```suggestion
            if (KEY.equals(fieldName)) {
```

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/service/validator/ServiceValidator.java
```diff
@@ -132,6 +138,75 @@ public void validate() {
...
+            for (Node expression : ((ListConstructorExpressionNode) keyFieldExpression).expressions()) {
+                if (expression.kind() != SyntaxKind.STRING_LITERAL) {
+                    addDiagnostic(CompilationDiagnostic.PROVIDE_AN_ARRAY_OF_STRING_LITERALS_FOR_KEY_FIELD,
+                                  expression.location(), KEY);
+                }
+            }
+            return;
+        }
+        addDiagnostic(CompilationDiagnostic.PROVIDE_A_STRING_LITERAL_OR_AN_ARRAY_OF_STRING_LITERALS_FOR_KEY_FIELD,
+                      keyFieldExpression.location(), KEY);
```
ThisaruGuruge, line 209: This function flow looks counter-intuitive. First `if` block adds a diagnostic and returns, while the second if block just returns. We may invert the second condition to add the diagnostic inside that block. 

When it is a list, aren't we doing the same check for list members? Can't we use a separate function to validate each expression and call it inside the loop?

commons/src/main/java/io/ballerina/stdlib/graphql/commons/utils/KeyDirectivesArgumentHolder.java
```diff
@@ -18,15 +18,17 @@
...
 package io.ballerina.stdlib.graphql.commons.utils;
 
+import java.io.Serializable;
 import java.util.List;
 
 /**
  * Hold the necessary arguments of GraphQL federation @key directive.
  */
-public class KeyDirectivesArgumentHolder {
+public class KeyDirectivesArgumentHolder implements Serializable {
```
ThisaruGuruge, line 27: This should be added to GraalVM configs since we serialize this class.

ThisaruGuruge:
We should update and release the graphql-tools with this change.

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/service/validator/ServiceValidator.java
MohamedSabthar: The trim method removes the leading or trailing minutiae.

docs/spec/spec.md
```diff
@@ -966,6 +966,8 @@ In Ballerina, `distinct` `service` objects can be used to define GraphQL interfa
 
 Non-distinct `service` objects and `service` classes can not be used to define or implement GraphQL interfaces.
 
+>**Note**: Recognition of `service` `object`s and `service` `class`xes as GraphQL interfaces and implementations relies on their definition within the same module as the GraphQL service.
```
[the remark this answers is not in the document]
MohamedSabthar: Updated

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/diagnostics/DiagnosticMessage.java
MohamedSabthar: Updated

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/service/validator/ServiceValidator.java
MohamedSabthar: Updated

commons/src/main/java/io/ballerina/stdlib/graphql/commons/utils/KeyDirectivesArgumentHolder.java
MohamedSabthar: Added

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/service/validator/ServiceValidator.java
MohamedSabthar: If we invert the second 'if' block, it will always result in a warning for a list constructor expression. We could modify that 'if' block to check both kinds (not equals to 'STRING_LITERAL' and 'LIST_CONSTRUCTOR'), move the warning on line 208 inside this block, and then return. This way, both solutions will yield the same result.

ThisaruGuruge: Then we should check for the list first. 

My understanding is, at the end of the day, what we want to check is whether a given expression is valid or not. So, as per the single responsibility principle, one function should only do that. Then there's the key argument, which can be either an expression or a list. For a list, we should iterate and pass each expression to the aforementioned function. Am I missing something?

MohamedSabthar: Got the point now. Separate the concerns of iteration and validation with: [a02f5db](https://github.com/ballerina-platform/module-ballerina-graphql/pull/1579/commits/a02f5dbead0dd96f4a4984827d16674941a774df)

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/service/validator/ServiceValidator.java
```diff
@@ -132,6 +138,75 @@ public void validate() {
...
+        if (keyFieldExpression.kind() == SyntaxKind.LIST_CONSTRUCTOR) {
+            for (Node expression : ((ListConstructorExpressionNode) keyFieldExpression).expressions()) {
+                validateKeyFieldValues(expression);
+            }
+            return;
+        }
+        validateKeyFieldValues(keyFieldExpression);
+    }
+
+    public void validateKeyFieldValues(Node expression) {
```
ThisaruGuruge, line 205:
```suggestion
    public void validateKeyFieldValue(Node expression) {
```

docs/spec/spec.md
```diff
@@ -966,6 +966,8 @@ In Ballerina, `distinct` `service` objects can be used to define GraphQL interfa
 
 Non-distinct `service` objects and `service` classes can not be used to define or implement GraphQL interfaces.
 
+>**Note**: Recognition of `service` `object`s and `service` `class`es as GraphQL interfaces and implementations relies on their definition within the same module as the GraphQL service.
```
ThisaruGuruge, line 969: On second thoughts, this seems unclear. Shall we reword this? IMO, the spec (or any technical documentation for that matter) should be easy to read and understand. Using excessive adjectives and complex sentences (generated by generative AIs) beats that purpose.

```suggestion
>**Note**: In order to be recognized as GraphQL objects or interfaces, the Ballerina `service` `object`s and `service` `class`es should be defined within the same module as the GraphQL service.
```

docs/spec/spec.md
```diff
@@ -3183,6 +3185,8 @@ To fully define an entity within a Ballerina GraphQL subgraph, you must:
 2. Define the `key` field of the annotation to be the fields and subfields that contribute to the entity's primary key/keys.
 3. Define the `resolveReference` field of the annotation to be a function pointer to resolve the entity. If this field is set to `nil`, it indicates to the graph router that this subgraph does not define a reference resolver for this entity. For more details, see [ReferenceResolver](#10513-the-subgraphreferenceresolver-function-type).
 
+>**Note:** To ensure recognition as subgraph entities by the GraphQL module, GraphQL object types marked with `@subgraph:Entity` must be defined within the same Ballerina module as the GraphQL service; failure to comply results in non-recognition.
```
ThisaruGuruge, line 3188:
```suggestion
>**Note:** In order to be recognized as subgraph entities, GraphQL object types marked with `@subgraph:Entity` must be defined within the same Ballerina module as the GraphQL service.
```

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/service/validator/ServiceValidator.java
MohamedSabthar: Updated with [9ece05c](https://github.com/ballerina-platform/module-ballerina-graphql/pull/1579/commits/9ece05c8eb4201782cd221a4e5c0ef371bcaa03d)

docs/spec/spec.md
```diff
@@ -966,6 +966,8 @@ In Ballerina, `distinct` `service` objects can be used to define GraphQL interfa
 
 Non-distinct `service` objects and `service` classes can not be used to define or implement GraphQL interfaces.
 
+>**Note**: In order to be recognized as GraphQL objects or interfaces, the Ballerina `service` `object`s and `service` `class`es should be defined within the same module as the GraphQL service.
```
ThisaruGuruge, line 969: Sorry, mised this.

```suggestion
>**Note**: In order to be recognized as GraphQL objects or interfaces, the Ballerina `service` `object`s and `service` `class`es must be defined within the same module as the GraphQL service.
```

ThisaruGuruge (approved):

PR #1735, opened 2024-02-01
Introduce GraphQL Server-side Caching

DimuthuMadushan:
## Purpose

Fixes: [#3621](https://github.com/ballerina-platform/ballerina-library/issues/3621)
Fixes: [#6017](https://github.com/ballerina-platform/ballerina-library/issues/6017)

## Examples
```ballerina
import ballerina/graphql;

@graphql:ServiceConfig {
    cacheConfig: {
        enabled: true,
        maxAge: 50
    }
}
service /graphql on new graphql:Listener(9090) {
    private string name = "Ballerina GraphQL";
    private string 'type = "code first";

    resource function get name() returns string {
        return self.name;
    }

    resource function get 'type() returns string {
        return self.'type;
    }

    remote function updateName(graphql:Context context, string name) returns string|error {
        check context.evictCache("name");
        self.name = name
        return self.name;
    }
}
```

## Checklist
- [X] Linked to an issue
- [X] Updated the changelog
- [X] Added tests
- [X] Updated the spec
- [X] Checked native-image compatibility
- [X] No `commons` package changes (if there are any, please update the GraphQL version in [GraphQL tools](https://github.com/ballerina-platform/graphql-tools) and [Ballerina dev tools](https://github.com/ballerina-platform/ballerina-dev-tools))
- [X] No `compiler` package changes (if there are any, please update the GraphQL version in [Ballerina dev tools](https://github.com/ballerina-platform/ballerina-dev-tools))


ballerina/engine.bal
```diff
@@ -351,6 +383,35 @@ isolated class Engine {
         _ = resourcePath.pop();
     }
 
+    private isolated function getFieldValue(Context context,Field 'field, ResponseGenerator responseGenerator) returns any|error {
```
MohamedSabthar, line 386:
```suggestion
    private isolated function getFieldValue(Context context, Field 'field, ResponseGenerator responseGenerator) returns any|error {
```

ballerina/engine.bal
```diff
@@ -351,6 +383,35 @@ isolated class Engine {
...
+            });
+            foreach string key in keys {
+                _ = check (<cache:Cache>self.cache).invalidate(key);
+            }
+        }
+        return;
+    }
+
+    isolated function invalidateAll() returns error? {
+        if self.cache is cache:Cache {
```
Nuvindu, line 409:
```suggestion
         cache:Cache? cache = self.cache;
         if cache is cache:Cache {
             return cache.invalidateAll();
         }
         return;
```

Nuvindu, line 409:
```suggestion
         cache:Cache? cache = self.cache;
         if cache is cache:Cache {
             return cache.invalidateAll();
         }
         return;
```

ballerina/listener.bal
```diff
@@ -71,11 +71,13 @@ public class Listener {
...
         boolean introspection = getIntrospection(serviceConfig);
         boolean validation = getValidation(serviceConfig);
+        ServerCacheConfig? operationCacheConfig = getCacheConfig(serviceConfig);
+        ServerCacheConfig? fieldCacheConfig = getFieldCacheConfigFromServiceConfig(serviceConfig);
         Engine engine;
         if self.graphiql.enabled {
             check validateGraphiqlPath(self.graphiql.path);
             string gqlServiceBasePath = name is () ? "" : getBasePath(name);
-            engine = check new (schemaString, maxQueryDepth, s, interceptors, introspection, validation);
+            engine = check new (schemaString, maxQueryDepth, s, interceptors, introspection, validation, operationCacheConfig, fieldCacheConfig);
```
ThisaruGuruge, line 80: Check character limits.

ballerina/response_generator.bal
```diff
@@ -23,14 +23,18 @@ class ResponseGenerator {
     private final Context context;
     private (string|int)[] path;
     private final __Type fieldType;
+    private final ServerCacheConfig? cacheConfig;
+    private final string[] parentArgHashes;
 
     private final string functionNameGetFragmentFromService = "";
 
-    isolated function init(Engine engine, Context context, __Type fieldType, (string|int)[] path = []) {
+    isolated function init(Engine engine, Context context, __Type fieldType, (string|int)[] path = [], ServerCacheConfig? cacheConfig = (), string[] parentArgHashes = []) {
```
ThisaruGuruge, line 31: Check the character limit. Fix similar places.

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/ServiceDeclarationAnalysisTask.java
```diff
@@ -64,5 +64,6 @@ public void perform(SyntaxNodeAnalysisContext context) {
         Schema schema = generateSchema(context, interfaceEntityFinder, node, description);
         DocumentId documentId = context.documentId();
         addToModifierContextMap(documentId, node, schema);
+        addToModifierContextMap(documentId, node, serviceValidator.getCacheConfigContext());
```
ThisaruGuruge, line 67: I think we have a local variable declaration analysis task as well. Shouldn't we include this there?

ThisaruGuruge (commented):
We need to bump the version to next minor version.

ThisaruGuruge:
Code coverage seems to be just touching 80%. Is there a way to increase the coverage?

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/ServiceDeclarationAnalysisTask.java
ThisaruGuruge: It is there, missed it.

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/schema/generator/GraphqlSourceModifier.java
```diff
@@ -470,6 +494,41 @@ private SpecificFieldNode getSchemaStringFieldNode(String schemaString) {
         return NodeFactory.createSpecificFieldNode(null, fieldName, colon, fieldValue);
     }
 
+    private SpecificFieldNode getFieldCacheConfigNode(CacheConfigContext cacheConfigContext) {
+        Node fieldName = NodeFactory.createIdentifierToken(FIELD_CACHE_CONFIG_FIELD);
```
ThisaruGuruge, line 498: Are we checking whether the resource method has the `get` accessor? We should throw an error if the cache config is provided for a `subscribe` resource, right?

docs/spec/spec.md
```diff
@@ -104,9 +104,14 @@ The conforming implementation of the specification is released and included in t
         * 7.1.6 [Service Interceptors](#716-service-interceptors)
         * 7.1.7 [Introspection Configurations](#717-introspection-configurations)
         * 7.1.8 [Constraint Configurations](#718-constraint-configurations)
+        * 7.1.9 [Operation Cache Configurations](#719-operation-cache-configurations)
```
ThisaruGuruge, line 107: Shall we use the terms, `operation-level` and `field-level` caching?

```suggestion
        * 7.1.9 [Operation-level Cache Configurations](#719-operation-cache-configurations)
```

docs/spec/spec.md
```diff
@@ -104,9 +104,14 @@ The conforming implementation of the specification is released and included in t
...
+            * 7.1.9.1 [The `enabled` Field](#7191-the-enabled-field)
+            * 7.1.9.2 [The `maxAge` Field](#7192-the-maxage-field)
+            * 7.1.9.3 [The `maxSize` Field](#7193-the-maxsize-field)
     * 7.2 [Resource Configuration](#72-resource-configuration)
         * 7.2.1 [Field Interceptors](#721-field-interceptors)
         * 7.2.2 [Prefetch Method Name Configuration](#722-prefetch-method-name-configuration)
+        * 7.2.3 [Field Cache Configurations](#723-field-cache-configuration)
```
ThisaruGuruge, line 114:
```suggestion
        * 7.2.3 [Field-level Cache Configurations](#723-field-cache-configuration)
```

docs/spec/spec.md
```diff
@@ -1760,6 +1774,45 @@ service on new graphql:Listener(9090) {
 }
 ```
 
+#### 7.1.9 Operation Cache Configurations
+
+The `cacheConfig` field is used to provide the GraphQL operation cache configuration to enable the caching for `query` operations.
```
ThisaruGuruge, line 1779: Check similar places.

```suggestion
The `cacheConfig` field is used to provide the GraphQL operation-level cache configuration to enable the caching for `query` operations.
```

docs/spec/spec.md
```diff
@@ -1760,6 +1774,45 @@ service on new graphql:Listener(9090) {
...
+
+###### Example: Operation Cache Configurations
+
+```ballerina
+@graphql:ServiceConfig {
+    cacheConfig: {
+        enabled: true,
+        maxAge: 60,
+        maxSize: 120
```
ThisaruGuruge, line 1788: These are all default values. Shall we use other values? We may move the enable cache with default values to the top, and put an example of overriding default values next.

docs/spec/spec.md
```diff
@@ -3647,3 +3746,141 @@ isolated function followersLoaderFunction(readonly & anydata[] ids) returns Foll
...
 
 The above example utilizes three DataLoader instances: `postsLoader`, `rePostsLoader`, and `followersLoader`. These DataLoaders are associated with the batch load functions `postsLoaderFunction`, `rePostsLoaderFunction`, and `followersLoaderFunction`. The 'post' field in the example utilizes the `postsLoader` and `rePostsLoader` DataLoaders, while the 'followers' field utilizes the `followersLoader` DataLoader. This demonstrates how different fields can utilize specific DataLoaders to efficiently load and retrieve related data in GraphQL resolvers.
+
+### 10.7 Caching
+
+This section describes the caching mechanisms in the Ballerina GraphQL module.
+
+#### 10.7.1 Server-side Caching
+
+The Ballerina GraphQL module offers built-in server-side caching for GraphQL `query` operations. The caching operates as in-memory caching, implemented using the Ballerina cache module. The GraphQL module generates cache keys based on the arguments and the path. In server-side caching, the `errors` and `null` values are skipped when caching. There are two different ways called `operation caching` and `field caching` to enable server-side caching.
```
ThisaruGuruge, line 3756: How does the pluggable cache work here?

docs/spec/spec.md
```diff
@@ -1760,6 +1774,45 @@ service on new graphql:Listener(9090) {
...
+```ballerina
+@graphql:ServiceConfig {
+    cacheConfig: {}
+}
+service on new graphql:Listener(9090) {
+    // ...
+}
+```
+
+##### 7.1.9.1 The `enabled` Field
```
ThisaruGuruge, line 1807: Use the markdown linter. 

```suggestion
##### 7.1.9.1 The `enabled` Field

```

ballerina/engine.bal
```diff
@@ -259,20 +283,28 @@ isolated class Engine {
...
                 return check validateInterceptorReturnValue(fieldType, result, interceptorName);
             }
             any fieldValue;
-            if 'field.getOperationType() == parser:OPERATION_QUERY {
-                fieldValue = check self.resolveResourceMethod(context, 'field, responseGenerator);
-            } else if 'field.getOperationType() == parser:OPERATION_MUTATION {
-                fieldValue = check self.resolveRemoteMethod(context, 'field, responseGenerator);
+            if 'field.getOperationType() == parser:OPERATION_QUERY && 'field.isCacheEnabled() {
+                string cacheKey = 'field.getCacheKey();
+                any|error cachedValue = self.getFromCache(cacheKey);
```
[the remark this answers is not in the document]
DimuthuMadushan: The field value is `any` type since there are service objects.

docs/spec/spec.md
DimuthuMadushan: Since we are yet implement the distributed caching, it's not added to the spec.

ballerina/engine.bal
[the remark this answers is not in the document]
ThisaruGuruge: We don't need to cache the service objects. We should cache the values. Caching objects defeats the purpose of caching as we still have to execute the resource methods, which will result in DB calls, etc.

docs/spec/spec.md
ThisaruGuruge: I think it is quite straightforward. We have to have a separate field to provide `cache:Cache` object with a custom implementation. Will it take more time to implement this? If that's not the case, I'd add that feature here as well.

ballerina/engine.bal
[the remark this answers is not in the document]
DimuthuMadushan: Service object is cached to reduce the execution of resource that return the service object. Since the resources are also cached, do not need to execute its resource method after cached.

docs/spec/spec.md
DimuthuMadushan: But we focused on in-memory caching with the proposal, which added distributed caching as a non-goal. It may takes time and we can handle it separately.

ballerina/engine.bal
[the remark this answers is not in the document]
ThisaruGuruge: That makes sense.

ballerina/field.bal
```diff
@@ -160,4 +178,33 @@ public class Field {
...
+        return self.parentArgHashes;
+    }
+
+    private isolated function generateCacheKey() returns string {
+        string resourcePath = "";
+        foreach string|int path in self.path {
+            resourcePath += string `${path}.`;
+        }
+        string hash = generateArgHash(self.internalNode.getArguments(), self.parentArgHashes);
+        return string `${resourcePath}${hash}`;
```
ThisaruGuruge, line 208: Why do we need `${resourcePath}${hash}` for the cache key? Isn't just `hash` is enough?

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/CacheConfigContext.java
```diff
@@ -0,0 +1,33 @@
+package io.ballerina.stdlib.graphql.compiler;
+
+public class CacheConfigContext {
+    private Boolean enabled;
+    private int maxSize;
+
+    public CacheConfigContext(Boolean enabled, int maxSize) {
```
ThisaruGuruge, line 7: Do we need these methods to be public?

docs/spec/spec.md
```diff
@@ -3647,3 +3749,141 @@ isolated function followersLoaderFunction(readonly & anydata[] ids) returns Foll
...
+        check context.invalidate("name");
+        self.name = name
+        return self.name;
+    }
+}
+```
+
+In this example, caching is enabled at the operation level. Therefore, the field `name` and `type` will be cached. When updating the name with a mutation, the cached values become invalid. Hence, the `invalidate` function can be used to evict the existing cache values.
+
+###### Example: Field-level Cache Enabling and Eviction
```
ThisaruGuruge, line 3822: Let's add another example with overriding top-level cache configurations using resource-level cache config value.

docs/spec/spec.md
```diff
@@ -3647,3 +3749,141 @@ isolated function followersLoaderFunction(readonly & anydata[] ids) returns Foll
...
+        cacheConfig: {
+            enabled: true,
+            maxAge 20
+        }
+    }
+    isolated resource function get age() returns int {
+        return self.age;
+    }
+
+    isolated resource function get isMarried() returns boolean {
```
ThisaruGuruge, line 3883: Maybe we can use another cache config here to disable caching for this field. (Let's use comments to describe the behavior as well).

docs/spec/spec.md
```diff
@@ -3647,3 +3749,141 @@ isolated function followersLoaderFunction(readonly & anydata[] ids) returns Foll
...
+
+##### 10.7.1.1 Operation-level Caching
+
+Operation-level caching can be used to cache the entire operation, and this can be enabled by providing the [operation cache configurations](#719-operation-level-cache-configurations). Once enabled, the GraphQL server initiates caching for all subfields of `query` operations. The fields requested through query operations will be cached based on the specified cache configurations
+
+##### 10.7.1.2 Field-level Caching
+
+The GraphQL field-level caching can be enabled only for a specific field. This can be done by providing the [field cache configurations](#723-field-level-cache-configuration). Once the field-level caching is enabled for a field, it will be applied to the sub-fields of that field.
```
ThisaruGuruge, line 3767: Let's mention here that field-level cache configs can be used to override the configs from the parents.

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/CacheConfigContext.java
DimuthuMadushan: yes

ballerina/field.bal
DimuthuMadushan: The resource path is used to find the cache entries when manually evicting as mentioned in the proposal.

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/CacheConfigContext.java
ThisaruGuruge: Who is accessing them from the outside? As it seems these can be package private.

ballerina/field.bal
ThisaruGuruge: Even when manually evicting, we can generate the hash and check for the key, right? 
Using the `resource path` + `hash` beats the purpose of hashing. We used hash to reduce the size of the keys. But now we are creating even longer keys.

docs/spec/spec.md
```diff
@@ -1820,6 +1876,28 @@ service on new graphql:Listener(9090) {
 }
 ```
 
+#### 7.2.3 Field-level Cache Configuration
+
+The `cacheConfig` field is used to provide the field-level cache configs. The fields are as same as the operation cache configs.
```
ThisaruGuruge, line 1881: Let's add an anchor here as well. Better to link to the specific section, instead of linking to the top-level cache topic.

docs/spec/spec.md
```diff
@@ -1760,6 +1774,48 @@ service on new graphql:Listener(9090) {
 }
 ```
 
+#### 7.1.9 Operation-level Cache Configurations
+
+The `cacheConfig` field is used to provide the operation-level cache configuration to enable the [GraphQL caching](#107-caching) for `query` operations.
```
ThisaruGuruge, line 1779:
```suggestion
The `cacheConfig` field is used to provide the operation-level cache configuration to enable the [GraphQL caching](#10711-operation-level-caching) for `query` operations.
```

native/src/main/java/io/ballerina/stdlib/graphql/runtime/engine/ArgumentHandler.java
```diff
@@ -516,7 +516,7 @@ private Object[] getArgumentsForMethod() {
         return result;
     }
 
-    private static Type getEffectiveType(IntersectionType intersectionType) {
+    public static Type getEffectiveType(IntersectionType intersectionType) {
```
ThisaruGuruge, line 519: No need to make this public, right?

```suggestion
    static Type getEffectiveType(IntersectionType intersectionType) {
```

native/src/main/java/io/ballerina/stdlib/graphql/runtime/engine/Engine.java
```diff
@@ -327,4 +330,28 @@ public static void executePrefetchMethod(Environment environment, BObject contex
...
+
+    public static boolean hasRecordReturnType(BObject serviceObject, BArray path) {
+        ResourceMethodType resourceMethod = (ResourceMethodType) getResourceMethod(serviceObject, path);
+        if (resourceMethod == null) {
+            return false;
+        }
+        return isRecordReturnType(resourceMethod.getType().getReturnType());
+    }
+
+    public static boolean isRecordReturnType(Type returnType) {
```
ThisaruGuruge, line 342: Should this be public?

ballerina/field.bal
DimuthuMadushan: When manually evicting, how can we generate the cache key related to a specific cache entry without knowing its arguments(if the cached resource function has arguments)?

compiler-plugin/src/main/java/io/ballerina/stdlib/graphql/compiler/CacheConfigContext.java
DimuthuMadushan: It used inside the service validator.

native/src/main/java/io/ballerina/stdlib/graphql/runtime/engine/Engine.java
DimuthuMadushan: This method called from ballerina side.

docs/spec/spec.md
```diff
@@ -1760,6 +1774,48 @@ service on new graphql:Listener(9090) {
...
+    // ...
+}
+```
+
+###### Example: Operation-level Cache Configurations
+
+```ballerina
+@graphql:ServiceConfig {
+    cacheConfig: {
+        enabled: false
```
ThisaruGuruge, line 1797: `enable: false` does not make sense here, right?

docs/spec/spec.md
```diff
@@ -2698,6 +2776,30 @@ If the specified key does not exist in the context, the `getDataLoader()` method
 dataloader:DataLoader authorLoader = context.getDataLoader("authorLoader");
 ```
 
+##### 10.1.1.6 Evict Cache from Context
+
+To evict a cache entry from the server-side caching, the `invalidate` method can be used. It requires one parameter:
```
ThisaruGuruge, line 2781:
```suggestion
The `invalidate()` method can be used to evict cache entries from the cache that are related to a particular field. It requires one parameter:
```

ThisaruGuruge (approved):

ThisaruGuruge (approved):

PR #1822, opened 2024-04-10
Bump the Dependency Versions

ThisaruGuruge:
## Purpose

$Subject


dilanSachi:
We need to change distribution version in the Ballerina.toml right?

ThisaruGuruge:
> We need to change distribution version in the Ballerina.toml right?

Missed this, will update.

dilanSachi (approved):

PR #2067, opened 2024-10-15
[2201.9.x] Introduce Parallel Execution for the GraphQL Resolvers

MohamedSabthar:
## Purpose

Fixes: https://github.com/ballerina-platform/ballerina-library/issues/4122

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [ ] Updated the spec
- [x] No `commons` package changes (if there are any, please update the GraphQL version in [GraphQL tools](https://github.com/ballerina-platform/graphql-tools) and [Ballerina dev tools](https://github.com/ballerina-platform/ballerina-dev-tools))
- [x] No `compiler` package changes (if there are any, please update the GraphQL version in [Ballerina dev tools](https://github.com/ballerina-platform/ballerina-dev-tools))

### DIFF
--- a/ballerina/context.bal
+++ b/ballerina/context.bal
@@ -43,6 +43,9 @@ public isolated class Context {
     }
 
     # Retrieves a value using the given key from the GraphQL context.
+    # ```ballerina
+    # string userId = check context.get("userId").ensureType();  
+    # ```
     #
     # + key - The key corresponding to the required value
     # + return - The value if the key is present in the context, a `graphql:Error` otherwise
@@ -55,6 +58,9 @@ public isolated class Context {
     }
 
     # Removes a value using the given key from the GraphQL context.
+    # ```ballerina
+    # string userId = check context.remove("userId").ensureType();  
+    # ```
     #
     # + key - The key corresponding to the value to be removed
     # + return - The value if the key is present in the context, a `graphql:Error` otherwise
@@ -67,6 +73,10 @@ public isolated class Context {
     }
 
     # Register a given DataLoader instance for a given key in the GraphQL context.
+    # ```ballerina
+    # dataloader:DataLoader userDataLoader = new dataloader:DefaultDataLoader(batchUsers);
+    # check context.registerDataLoader("user", userDataLoader);
+    # ```
     #
     # + key - The key for the DataLoader to be registered
     # + dataloader - The DataLoader instance to be registered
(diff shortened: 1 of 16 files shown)


native/src/main/java/io/ballerina/stdlib/graphql/runtime/engine/ExecutorVisitor.java
```diff
@@ -0,0 +1,65 @@
...
+import static io.ballerina.stdlib.graphql.runtime.utils.ModuleUtils.getModule;
+
+/**
+ * This class provides native implementations of the Ballerina ExecutorVisitor class.
+ */
+public class ExecutorVisitor {
+    private static final String DATA_MAP = "dataMap";
+    private static final String DATA_RECORD_NAME = "Data";
+    private static final Object NullObject = new Object();
+    private final ConcurrentHashMap<BString, Object> dataMap = new ConcurrentHashMap<>();
```
DimuthuMadushan, line 37: When initializing the ConcurrentHashMap, we can configure the concurrency level.

* `concurrencyLevel`: The estimated number of concurrently updating threads. The implementation may use this value as a sizing hint. (Default: 16)

Is it okay to use the default value, or do we need to provide a way to configure it? WDYT @ThisaruGuruge ?

ThisaruGuruge: AFAIK, the `concurrencyLevel` is not significant in the later Java versions. So, let's not worry about this for now. We can improve this later if we face any issues.

ballerina/context.bal
```diff
@@ -22,87 +22,65 @@ import ballerina/lang.value;
...
-                self.attributes['key] = value.clone();
-            } else {
-                self.attributes['key] = value;
-            }
-        }
+        self.setAttribute('key, value is value:Cloneable ? value.clone() : value);
     }
 
     # Retrieves a value using the given key from the GraphQL context.
     #
```
ThisaruGuruge, line 46: Shall we include a sample code here, similar to what we have in langlib functions? Something like: (Please check the syntax, it might be wrong)

```suggestion
# ```
# string userId = check context.get("userId").ensureType();
# ```
```

ThisaruGuruge (approved):

PR #2184, opened 2025-03-17
Fix Fragments Returning null Value for non-null Types 

DimuthuMadushan:
## Purpose

Fixes: [#7689](https://github.com/ballerina-platform/ballerina-library/issues/7689)

## Checklist
- [X] Linked to an issue
- [X] Updated the changelog
- [X] Added tests
- [ ] ~Updated the spec~
- [X] Checked native-image compatibility
- [X] No `commons` package changes (if there are any, please update the GraphQL version in [GraphQL tools](https://github.com/ballerina-platform/graphql-tools) and [Ballerina dev tools](https://github.com/ballerina-platform/ballerina-dev-tools))
- [X] No `compiler` package changes (if there are any, please update the GraphQL version in [Ballerina dev tools](https://github.com/ballerina-platform/ballerina-dev-tools))


ballerina/response_formatter.bal
```diff
@@ -82,18 +85,28 @@ class ResponseFormatter {
...
+                                           string onType) returns Data? {
         foreach parser:SelectionNode selection in fragmentNode.getSelections() {
             if selection is parser:FragmentNode {
-                self.coerceFragmentValues(data, result, selection, parentType, selection.getOnType());
+                Data? fieldResult = self.coerceFragmentValues(data, result, selection, parentType, selection.getOnType());
+                if fieldResult == () {
+                    return;
+                }
             } else if selection is parser:FieldNode {
                 if data.hasKey(selection.getAlias()) || data.hasKey(selection.getName()) {
```
ThisaruGuruge, line 96: This would be enough right? `getAlias()` should return the name, if the alias does not exist. 

```suggestion
                if data.hasKey(selection.getAlias()) {
```

ThisaruGuruge (approved):
Approved with a minor comment.

ballerina/response_formatter.bal
DimuthuMadushan: This can affect a service that uses interceptors to update the `data`.

DimuthuMadushan:
```ballerina
readonly service class RecordInterceptor1 {
    *graphql:Interceptor;

    isolated remote function execute(graphql:Context context, graphql:Field 'field) returns anydata|error {
        var result = context.resolve('field);
        if result is record {} && 'field.getName() != "contact" {
            return {
                name: "Rubeus Hagrid",
                age: 70,
                address: {number: "103", street: "Mould-on-the-Wold", city: "London"}
            };
        }
        return result;
    }
}
```
If we use an interceptor like the one above, it completely updates the `Data` record with the original field names, even if it previously had aliases. When the response formatter checks for aliases, it cannot find them since the data record includes only the original field names.

ThisaruGuruge: In this case, doesn't it break the response by replacing the alias? I think this is incorrect behavior from the interceptor, as it should check the aliases before re-assigning the value. But let's keep this as it is for now.

PR #2534, opened 2026-07-20
feat: add subscription support for the client

### LINKED ISSUE #8915: GraphQL Client Subscription Support
### Area

None

### Problem

Ballerina GraphQL client does not support GraphQL subscription, which is a major missing point.

### Proposed Solution

Add support for the Subscription support for the Ballerina GraphQL client.

### Alternatives

_No response_

### Version

_No response_

ThisaruGuruge:
## Purpose

Fixes: https://github.com/ballerina-platform/ballerina-library/issues/8915

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [x] Updated the spec
- [x] No `commons` package changes (if there are any, please update the GraphQL version in [GraphQL tools](https://github.com/ballerina-platform/graphql-tools) and [Ballerina dev tools](https://github.com/ballerina-platform/ballerina-dev-tools))
- [x] No `compiler` package changes (if there are any, please update the GraphQL version in [Ballerina dev tools](https://github.com/ballerina-platform/ballerina-dev-tools))

## Summary

Adds GraphQL subscription support to the Ballerina GraphQL client.

## Key changes

- Adds `Client->subscribe()` with typed event streams over `graphql-transport-ws`.
- Adds subscription lifecycle management, multiplexing, stream closure, and client closure.
- Adds configurable reconnection and keep-alive handling.
- Adds `query()` and `mutate()` APIs and deprecates `executeWithType()`.
- Improves operation validation and preserves partial response data in `PayloadBindingError`.
- Adds server-side keep-alive configuration for GraphQL subscriptions.
- Adds extensive subscription, reconnection, authentication, error handling, and keep-alive tests.
- Updates documentation, changelog entries, examples, native runtime support, and dependency metadata.


ballerina/client_subscription_connection.bal
```diff
@@ -0,0 +1,517 @@
...
+        websocket:Error? result = connectionResult->writeMessage(subscribeMessage);
+        if result is websocket:Error {
+            _ = self.removeOperation(id);
+            self.unlockState();
+            return error SubscriptionError(string `Failed to send the subscribe message: ${result.message()}`,
+                    result, errors = ());
+        }
+        self.unlockState();
+        return;
+    }
```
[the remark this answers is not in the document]
ThisaruGuruge: Addressed in 8a1e7bc2: subscribe() now sets the connecting flag, releases the state lock across connect(), then re-acquires and re-validates (closed / superseding client) before storing the connection and sending the pending subscribe messages — mirroring reconnect()/handleConnectionFailure().

ballerina/client_subscription_stream.bal
```diff
@@ -0,0 +1,94 @@
...
+                ClientError? unsubscribeResult = self.connection.unsubscribe(self.id);
+                if unsubscribeResult is ClientError {
+                    logError("Failed to unsubscribe from the subscription operation", unsubscribeResult);
+                }
+                return response;
+            }
+            return {value: response};
+        }
+        return error SubscriptionError(INVALID_SUBSCRIPTION_MESSAGE, errors = ());
+    }
```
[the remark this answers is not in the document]
ThisaruGuruge: Addressed in 8a1e7bc2: replaced the isClosed() short-circuit with a local `terminated` flag set only after a terminal value is actually dequeued, so buffered payloads are drained before termination is observed.

ballerina/common_utils.bal
```diff
@@ -454,8 +424,15 @@ isolated function performDataBindingWithErrors(typedesc<GenericResponseWithError
...
-        return error PayloadBindingError(UNABLE_TO_PERFORM_DATA_BINDING, e, errors = errorDetails);
+        ErrorDetail[]|error errorDetails = responseMap.get("errors").cloneWithType();
+        if errorDetails is error {
+            return error PayloadBindingError(UNABLE_TO_PERFORM_DATA_BINDING, e, errors = ());
+        }
+        json? data = responseMap.hasKey("data") ? responseMap.get("data") : ();
+        map<json>? extensions = responseMap.hasKey("extensions") ?
+            (responseMap.get("extensions") is () ? () : <map<json>>responseMap.get("extensions")) : ();
+        return error PayloadBindingError(UNABLE_TO_PERFORM_DATA_BINDING, e,
+                data = data, errors = errorDetails, extensions = extensions);
```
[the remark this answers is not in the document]
ThisaruGuruge: Addressed in 8a1e7bc2: replaced the `<map<json>>` cast with an `is map<json>` type test, so a non-map `extensions` value falls back to `()` instead of panicking.

native/src/main/java/io/ballerina/stdlib/graphql/runtime/client/Client.java
```diff
@@ -41,12 +46,42 @@ public static Object execute(Environment env, BObject client, BString document,
...
+        paramFeed[3] = operationName;
+        paramFeed[4] = id;
+        Object result = invokeClientMethod(env, client, "processSubscribe", paramFeed);
+        if (result instanceof BError) {
+            return result;
+        }
+        StreamType streamType = TypeCreator.createStreamType(targetType.getDescribingType(),
+                PredefinedTypes.TYPE_NULL);
+        return ValueCreator.createStreamValue(streamType, (BObject) result);
     }
```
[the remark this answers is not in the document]
ThisaruGuruge: Keeping `PredefinedTypes.TYPE_NULL` intentionally. The declared return type `stream<targetType, ClientError?>` governs static typing at call sites, and `stream<T, ()>` is a subtype of `stream<T, ClientError?>` (completion type is covariant), so the returned value is type-safe and error completion still works — `SubscriptionStreamGenerator.next()` returns a `ClientError` on failure and consumers handle it (covered by `testInvalidSubscription`). This matches the platform idiom: `sql:Client->query()` declares `stream<rowType, sql:Error?>` but constructs the runtime stream with `TYPE_NULL` (QueryProcessor.java). Building the precise `ClientError|()` union would add interop for no behavioral change and diverge from sql.

ballerina/client.bal
```diff
@@ -14,72 +14,136 @@
...
-        request.setPayload(graphqlPayload);
-        json|http:ClientError httpResponse = self.httpClient->post("", request, headers = headers);
-
-        if httpResponse is http:ClientError {
-            return handleHttpClientErrorResponse(httpResponse);
+    private isolated function processQuery(typedesc<GenericResponseWithErrors|record {}|json> targetType,
+            string document, map<anydata>? variables, string? operationName, map<string|string[]>? headers)
+            returns GenericResponseWithErrors|record {}|json|ClientError {
+        if self.subscriptionConnection.isClosed() {
+            return error ClientError(CLIENT_ALREADY_CLOSED_MESSAGE);
```
DimuthuMadushan: Why do we need to check the subscription connection status here?

ThisaruGuruge: We are keeping track of the connections in a map. If a given connection is closed (using the `close()` method), we should stop the subsequent calls.

ballerina/client_subscription_connection.bal
```diff
@@ -0,0 +1,724 @@
...
+            if message is websocket:Error {
+                if keepAliveMonitor is KeepAliveMonitor {
+                    keepAliveMonitor.stop();
+                    self.keepAliveSignalQueue.enqueue(());
+                }
+                self.handleConnectionFailure(wsClient);
+                return;
+            }
+            if keepAliveMonitor is KeepAliveMonitor {
+                keepAliveMonitor.markPongReceived();
```
ThisaruGuruge: **Correctness bug:** `keepAliveMonitor.markPongReceived()` is called for every incoming message (line 253), not just `WS_PONG` frames — the `getWsMessageType(message) == WS_PONG` check only happens afterward, to decide whether to skip dispatch. This means any subscription traffic (`next`/`error`/`complete`/`ping`) resets the missed-probe counter in `runKeepAlive`, so a server that stops answering `ping` specifically (e.g. a stuck control-plane handler) but keeps pushing `next" frames for other active subscriptions will never accumulate `KEEPALIVE_MAX_MISSED_PROBES" and `handleKeepAliveTimeout` will never fire. This also contradicts the class doc comment above (~line 625) which states the dispatcher "sets it upon reading a `pong`."

Suggest gating the `markPongReceived()` call on the message actually being a `WS_PONG`.

ballerina/client.bal
```diff
@@ -94,26 +205,54 @@ public isolated client class Client {
...
+        'class: "io.ballerina.stdlib.graphql.runtime.client.Client",
         name: "execute"
     } external;
 
     private isolated function processExecute(typedesc<GenericResponseWithErrors|record{}|json> targetType,
                                              string document, map<anydata>? variables, string? operationName,
                                              map<string|string[]>? headers)
                                              returns GenericResponseWithErrors|record{}|json|ClientError {
+        if self.subscriptionConnection.isClosed() {
+            return error ClientError(CLIENT_ALREADY_CLOSED_MESSAGE);
```
ThisaruGuruge, line 236: **Correctness bug (existing code re-exposed by this PR):** `processExecute` here (and `processQuery`/`processMutate`/`processSubscribe`) all start with `self.subscriptionConnection.isClosed()` guard before doing any work. However `processExecuteWithType` (lines 173-194, unchanged by this diff — it's the restored deprecated implementation) has no such guard and still hand-rolls its own HTTP request instead of calling `sendGraphqlRequest`.

Now that this PR introduces the notion of a closeable client (`client->close()` closing `subscriptionConnection`), calling the deprecated `executeWithType()` after `close()` will silently attempt an HTTP request instead of failing fast with `CLIENT_ALREADY_CLOSED_MESSAGE` like every other entry point (including the sibling deprecated `execute()`, right here). Since the PR restores `executeWithType()` specifically so it keeps working, this inconsistency is worth fixing: add the same `isClosed()` check (and ideally route through `sendGraphqlRequest`) in `processExecuteWithType`.

ballerina/common_utils.bal
```diff
@@ -435,29 +435,30 @@ isolated function performDataBinding(typedesc<GenericResponse|record{}|json> tar
     return error RequestError("GraphQL Client Error, Invalid binding type.");
 }
 
-isolated function performDataBindingWithErrors(typedesc<GenericResponseWithErrors|record{}|json> targetType,
+isolated function performDataBindingWithErrors(typedesc<GenericResponseWithErrors|record{}> targetType,
```
ThisaruGuruge, line 459: **Behavioral inconsistency between the deprecated and new APIs:** `performDataBindingWithErrors` (used by `query()`/`mutate()`/`execute()` via `processExecute"/`executeGraphqlDocument`) always attempts `cloneWithType(targetType)` first — so a response with partial `data` alongside `errors` can bind successfully into `GenericResponseWithErrors`, only falling back to `PayloadBindingError` if the clone itself fails.

The deprecated `executeWithType()` path (`processExecuteWithType` -> `handleGraphqlErrorResponse`, `client.bal` lines 189-190) instead short-circuits to a `ServerError` the instant the response JSON has an `errors` key, without ever attempting to bind the (possibly-present) partial `data`.

This means the exact same server response yields a successful, data-bound result via `query()`/`mutate()` but always an error via `executeWithType()`/`execute()` — a semantic difference callers migrating between the deprecated and new APIs won't expect, and it isn't called out in either method's deprecation notice.

native/src/main/java/io/ballerina/stdlib/graphql/runtime/client/SubscriptionConnectionState.java
```diff
@@ -0,0 +1,183 @@
...
+
+    public static boolean isClosed(BObject connection) {
+        return getState(connection).closed.get();
+    }
+
+    /**
+     * Acquires the lock serializing the compound state transitions of the connection. Every acquisition must be
+     * paired with an {@code unlockState} call.
+     */
+    public static void lockState(Environment env, BObject connection) {
```
ThisaruGuruge, line 151: **Concurrency design risk:** `lockState`/`unlockState` rely on a single `Semaphore(1)` acquired and released manually (not RAII/try-finally) across the corresponding Ballerina call sites in `client_subscription_connection.bal` (e.g. `subscribe()`, `close()`, `reconnect()`, `handleConnectionFailure()`). If any native call made between a `lockState()` and its matching `unlockState()` panics unexpectedly (e.g. an unforeseen native exception in `registerOperation`/`removeOperation`/`sendPendingSubscribeMessages`), the Ballerina runtime propagates the panic and skips the remaining statements -- including the `unlockState()` call -- permanently leaking the single permit. Every subsequent `subscribe()`/`close()`/`reconnect()` call on that client would then block forever waiting to acquire a permit that's never released, deadlocking the connection for its remaining lifetime.

Worth double-checking that nothing between lock/unlock pairs can realistically panic, or wrapping the critical sections so the lock is always released regardless of how the block exits.

ballerina/client_utils.bal
```diff
@@ -0,0 +1,135 @@
...
+            if operation.getName() == operationName {
+                return operation;
+            }
+        }
+        return {message: string `Unknown operation named "${operationName}".`, locations: []};
+    }
+    if operations.length() == 1 {
+        return operations[0];
+    }
+    return {message: MULTIPLE_OPERATIONS_MESSAGE, locations: []};
```
ThisaruGuruge: **Reuse:** `selectOperation` re-implements the "select the operation from a document by name / fall back if there's exactly one" logic that `Engine.getOperation` (`engine.bal` lines 315-339) already has. It even introduces a new `MULTIPLE_OPERATIONS_MESSAGE` constant (`constants.bal` line 89) for a message `engine.bal` line 321 still hardcodes as a literal string (`"Must provide operation name if query contains multiple operations."`) -- identical text, two independently-maintained copies. A future wording tweak to one is easy to miss in the other.

Consider having the client path call (or share a helper with) `Engine.getOperation` instead of duplicating the lookup.

ballerina/client_subscription_connection.bal
```diff
@@ -0,0 +1,724 @@
...
+                if getWsMessageType(message) == WS_PONG {
+                    continue;
+                }
+            }
+            self.dispatchMessage(wsClient, message);
+        }
+    }
+
+    // Pings only when idle; tears down only after KEEPALIVE_MAX_MISSED_PROBES silent cycles.
+    isolated function runKeepAlive(websocket:Client wsClient, KeepAliveMonitor monitor) {
```
ThisaruGuruge: **Reuse/altitude:** This client-side keep-alive state machine (`runKeepAlive`/`keepAliveWait`/`KeepAliveMonitor`, missed-probe counting against `KEEPALIVE_MAX_MISSED_PROBES`) is a near line-for-line mirror of the new server-side keep-alive in `websocket_service.bal` (~lines 130-190) -- both comments even reference the same upstream issue (ballerina-library#8929). Maintaining two independent implementations of the same ping/pong liveness protocol means a fix or edge case discovered on one side (like the markPongReceived-on-any-message issue flagged elsewhere in this review) can easily not get applied to the other. Worth factoring the shared ping/pong/missed-probe state machine into one reusable piece both the client and the listener depend on.

ballerina/client_utils.bal
```diff
@@ -0,0 +1,135 @@
...
+        responseLimits: config.responseLimits,
+        secureSocket: config.secureSocket,
+        proxy: config.proxy,
+        validation: config.validation
+    };
+}
+
+isolated function validateOperationKind(string document, string? operationName,
+        parser:RootOperationType expectedKind) returns InvalidDocumentError? {
+    parser:Parser documentParser = new (document);
```
ThisaruGuruge, line 45: **Efficiency:** `validateOperationKind` fully parses the GraphQL document client-side (`new parser:Parser(document).parse()`) on every single `query()`/`mutate()`/`subscribe()` call, solely to check the operation kind/name, then discards the AST. The server then re-parses the identical document from scratch to actually execute it. This doubles parsing cost on every call, scaling with document size and call rate -- worth considering whether the kind/name check can reuse the same parse that's needed to build the request payload, or be deferred to the server's own validation.

native/src/main/java/io/ballerina/stdlib/graphql/runtime/client/SubscriptionConnectionState.java
```diff
@@ -0,0 +1,183 @@
...
+ * structures: the active subscription operations, the closed flag, the connecting flag, and the current WebSocket
+ * client. Compound state transitions are serialized through an ownerless lock exposed via the
+ * {@code lockState}/{@code unlockState} functions (a strand may resume on a different thread, so a thread-owned lock
+ * cannot be used).
+ */
+public final class SubscriptionConnectionState {
+    private static final String NATIVE_STATE_KEY = "graphql.client.subscription.connection.state";
+
+    private final ConcurrentHashMap<String, SubscriptionOperation> operations = new ConcurrentHashMap<>();
+    private final AtomicBoolean closed = new AtomicBoolean(false);
```
ThisaruGuruge: **Altitude:** Connection lifecycle is tracked as two independent `AtomicBoolean`s (`closed`, `connecting`) rather than one explicit state (e.g. an enum: DISCONNECTED/CONNECTING/CONNECTED/RECONNECTING/CLOSED). Every transition site in `client_subscription_connection.bal` has to manually re-check and re-set both flags under the lock, and combinations that shouldn't be meaningful (e.g. momentarily `closed=true` while `connecting=true`, guarded only by scattered re-checks like the ones in `subscribe()`/`reconnect()`) are representable rather than made unrepresentable by the type. A single state field would make the valid transitions explicit and remove the need to reason about flag combinations at each call site.

ballerina/errors.bal
```diff
@@ -37,14 +37,19 @@ public type HttpError distinct (RequestError & error<record {| anydata body; |}>
...
 # Represents GraphQL API response during GraphQL API server side errors.
 # # Deprecated
 # This error type will be removed along with the `executeWithType()` API
 @deprecated
 public type ServerError distinct (ClientError & error<record {| json? data?; ErrorDetail[] errors; map<json>? extensions?; |}>);
 
-# Represents client side data binding error.
-public type PayloadBindingError distinct (ClientError & error<record {| ErrorDetail[]? errors; |}>);
+# Represents errors occurring while establishing or executing a GraphQL subscription.
+public type SubscriptionError distinct (ClientError & error<record {| ErrorDetail[]? errors; |}>);
```
ThisaruGuruge, line 52: **Altitude:** `PayloadBindingError` (`data?`/`errors?`/`extensions?`), `ServerError` (`data?`/`errors` required/`extensions?`), and `SubscriptionError` (`errors?` only, no `data`/`extensions`) are three separate `distinct` branches under `ClientError` with overlapping but inconsistent shapes, rather than one generalized "carries GraphQL error details" shape all three implement. Consumers have to branch on the concrete error type instead of matching a shared shape to read `errors`/`data`/`extensions` uniformly, and the next new error source (subscription reconnection errors, etc.) is likely to repeat the pattern with yet another bespoke field set instead of fitting an existing one.

ballerina/client.bal
```diff
@@ -14,26 +14,137 @@
...
+    # + return - The data-bound response, or a `graphql:ClientError` if the execution fails
+    remote isolated function query(string document, map<anydata>? variables = (),
+            string? operationName = (), map<string|string[]>? headers = (),
+            typedesc<GenericResponseWithErrors|record {}> targetType = <>)
+            returns targetType|ClientError = @java:Method {
+        'class: "io.ballerina.stdlib.graphql.runtime.client.Client",
+        name: "query"
+    } external;
+
+    private isolated function processQuery(typedesc<GenericResponseWithErrors|record {}> targetType,
```
ThisaruGuruge, line 66: **Simplification:** `processQuery` and `processMutate` are identical apart from the `parser:OPERATION_QUERY` vs `parser:OPERATION_MUTATION` constant passed to `validateOperationKind` (compare with `processMutate` a few lines below). Minor, but a shared private helper parameterized on the expected operation kind would remove the duplicated closed-check/validate/execute sequence and keep future changes to that sequence in one place.

ballerina/client_subscription_connection.bal
ThisaruGuruge: This is intentional, not a bug: any incoming message counts as a liveness signal by design, not just an explicit `WS_PONG` frame. Receiving anything at all from the server (ordinary subscription traffic included) is itself proof the connection is alive, similar to how TCP keepalive treats any received data as proof of life -- so gating `markPongReceived()` on `WS_PONG` would actually narrow the liveness check, not fix it.

That said, the secondary point holds up: the class doc comment ("dispatcher sets it upon reading a `pong`") and the `markPongReceived`/`isPongReceived`/`resetPong` naming did describe a narrower behavior than what the code actually does, which is exactly the kind of mismatch that invited this finding. Fixed that in a follow-up commit: renamed to `markLivenessSignalReceived`/`isLivenessSignalReceived`/`resetLivenessSignal` and reworded the class doc comment to explicitly describe any-message-counts-as-liveness. No behavioral change, just aligning names/docs with the deliberate design.

native/src/main/java/io/ballerina/stdlib/graphql/runtime/client/SubscriptionConnectionState.java
ThisaruGuruge: Verified: walked every lock/unlock pair in `client_subscription_connection.bal` (`subscribe()`, `unsubscribe()`, `close()`, `reconnect()`, `handleConnectionFailure()`, `handleKeepAliveTimeout()`). The critical sections only ever call: simple native accessors on `SubscriptionConnectionState` (`ConcurrentHashMap`/`AtomicReference`/`AtomicBoolean` ops -- no allocation, casting, or indexing that can throw), and `websocket:Client` remote calls (`writeMessage`/`close`) whose failures surface as checked `websocket:Error` unions, not panics. There's no `checkpanic`, unsafe cast, or array indexing anywhere between a `lockState()`/`unlockState()` pair today, so there's no realistic panic path that would leak the permit right now.

Making the lock robust against *arbitrary future* panics would need either wrapping every critical section in a `trap`-guarded closure (non-trivial here since `isolated function` closures capturing `self` plus several locals across ~6 call sites would need careful isolation-safety review) or redesigning the lock so a single native call owns the whole critical section end-to-end -- both are real restructuring, not a targeted fix. Given the current code has no panic surface to protect against, I'm leaving this as-is for this pass rather than doing a speculative refactor, but it's a fair thing to revisit if a future critical section adds anything panic-prone.

ballerina/client_subscription_connection.bal
ThisaruGuruge: Confirmed -- this is a real, near line-for-line duplicate of `websocket_service.bal`'s `runKeepAlive`/`keepAliveWait` (the server side already even settled on the more accurate `activityReceived`/`markActivity` naming that this pass borrows for the client side too, see the reply on the other keep-alive thread).

Agreed it's worth unifying, but doing it properly means abstracting over a real difference between the two: the server side pings a `websocket:Caller` plus its own `SubscriptionHandler`/`activeConnections` bookkeeping and closes via `closeConnection`, while the client side pings a `websocket:Client` plus `MessageQueue`-based dispatch and closes via `handleConnectionFailure`/reconnection. A shared state machine would need a transport-agnostic interface over both, which is a bigger structural change than fits this fix-response pass. Tracking as a follow-up rather than folding it in here.

ballerina/client_utils.bal
ThisaruGuruge: Valid observation, but this is a deliberate tradeoff rather than an oversight: the client-side parse exists specifically so a wrong-operation-kind call (e.g. calling `query()` with a mutation document) fails immediately with a clear `InvalidDocumentError`, without a network round trip. Deferring the check to the server's own validation would mean removing that fast, network-free failure mode entirely (the server would instead reject over the wire, or in the worst case for `query()`/`mutate()` vs a mismatched kind, execute the wrong thing) -- that's an API/behavior change, not a cost-neutral optimization.

There's no clean way to reuse this parse for the actual request either: the client sends the raw document string as-is (see `getGraphqlPayload`), it never serializes the AST back out, so there's nothing downstream that could consume the already-built `parser:DocumentNode`. Leaving this as-is for this pass; happy to revisit if client-side kind validation is ever reconsidered as a feature.

native/src/main/java/io/ballerina/stdlib/graphql/runtime/client/SubscriptionConnectionState.java
ThisaruGuruge: Valid design point, but replacing the two `AtomicBoolean`s with a single state enum touches every transition site in `client_subscription_connection.bal` (`subscribe()`, `reconnect()`, `handleConnectionFailure()`, `handleKeepAliveTimeout()`, `close()`) plus the native accessors, and needs care to preserve the exact current semantics -- e.g. `closed` and `connecting` are read outside the lock in a couple of places (`subscribe()`'s pre-lock `isClosed()` fast path) while being read/written under the lock elsewhere, so an enum swap isn't a pure rename, it's a concurrency-sensitive rewrite of the whole state machine. That's more than a targeted fix warrants for a review-response pass. Tracking as a follow-up refactor rather than doing it here.

ballerina/errors.bal
ThisaruGuruge: Valid observation, but unifying these is a public API design decision, not a targeted fix. All three (`PayloadBindingError`, `ServerError`, `SubscriptionError`) are `public` error types; introducing a shared "carries GraphQL error details" shape (a common detail-record type or interface all three implement) would be an additive-or-breaking change to the module's public error surface that deserves its own design discussion, not something to slip in while responding to review comments. It's also worth noting `ServerError` is already `@deprecated` and slated for removal, and `PayloadBindingError`/`SubscriptionError` are brand new in this PR -- so reshaping now, before either has a single external consumer, might be the better time to do it, but as a deliberate follow-up rather than folded into this pass.

ballerina/client_utils.bal
ThisaruGuruge: Good catch. Fixed: extracted the shared selection logic into a single `selectOperation` function in `common_utils.bal` (used by both the client and `Engine.getOperation`), so there's one implementation and one error-message constant (`MULTIPLE_OPERATIONS_MESSAGE`) instead of two independently-maintained copies. `engine.bal`'s `getOperation` now just delegates to it.

ballerina/client.bal
ThisaruGuruge: Fixed: added the same `self.subscriptionConnection.isClosed()` guard to `processExecuteWithType`, and while there, also routed it through `sendGraphqlRequest` instead of hand-rolling the HTTP request, matching `processExecute`'s pattern -- so `executeWithType()` now fails fast with `CLIENT_ALREADY_CLOSED_MESSAGE` after `client->close()`, same as every other entry point.

ballerina/common_utils.bal
ThisaruGuruge: Fixed: `processExecuteWithType` now attempts `performDataBinding(targetType, httpResponse)` first when `errors` is present, and only falls back to `handleGraphqlErrorResponse` (the `ServerError`) if that binding itself fails -- mirroring `query()`/`mutate()`'s `performDataBindingWithErrors` behavior. A response with partial `data` alongside `errors` now binds successfully via `executeWithType()` too, instead of always short-circuiting to an error.

ballerina/client.bal
ThisaruGuruge: Fixed: extracted a shared private `processOperation(parser:RootOperationType expectedKind, ...)` helper holding the closed-check/validate/execute sequence, and `processQuery`/`processMutate` now just call it with `parser:OPERATION_QUERY`/`parser:OPERATION_MUTATION` respectively.

ballerina/client_subscription_connection.bal
ThisaruGuruge: Follow-up: reversing my conclusion above after checking this against the actual `graphql-transport-ws` ecosystem convention rather than treating it as a plain correctness question in isolation.

The protocol spec (PROTOCOL.md) itself is silent on this -- implementation-defined either way. But the reference implementation, `enisdenjo/graphql-ws` (the library that effectively defines this protocol in practice), strictly gates its keep-alive timer reset on receiving an actual `pong`: in its `client.ts`, only `message.type === 'pong'` resets the timer -- `next`/`error`/`complete` do not. There's no "any message counts" logic in the reference client.

So the original suggested fix here was correct after all, and I'm implementing it now: `markPongReceived()` is gated on `getWsMessageType(message) == WS_PONG` again, and I've reverted the `markLivenessSignalReceived`/etc. naming back to `markPongReceived`/`isPongReceived`/`resetPong`, with the class doc comment updated to explicitly describe the pong-specific convention and cite `enisdenjo/graphql-ws` as the reason. Also found and fixed the same issue on the server side (`websocket_service.bal`): `onComplete`/`onSubscribe`/`onPingMessage` were all marking activity on non-pong traffic too -- now only `onPongMessage` does, for the same reasoning, keeping both sides of the protocol implementation consistent.

native/src/main/java/io/ballerina/stdlib/graphql/runtime/client/SubscriptionConnectionState.java
ThisaruGuruge: Update: implemented this properly in 5a37a62c. Every `lockState()`/`unlockState()` pair in `client_subscription_connection.bal` (`subscribe()`, `unsubscribe()`, `close()`, `reconnect()`, `handleConnectionFailure()`, `handleKeepAliveTimeout()`) now wraps its critical section in `trap` and calls `unlockState()` unconditionally right after, re-panicking once unlocked if the trapped value was an actual panic. Each critical section was moved into a private `...Locked` helper returning a plain (non-`error`) outcome type -- `SubscribeStartOutcome`, `SubscribeCompletion`, `ConnectionFailureOutcome`, `KeepAliveTimeoutOutcome`, `ReconnectAttemptOutcome` -- so `trapped is error` unambiguously means a caught panic, never a legitimate `ClientError`/`SubscriptionError` return. The calling function still performs any post-unlock action (closing the WebSocket, starting reconnection) exactly where it did before, preserving the original before/after-unlock ordering. No behavior change on the non-panicking path -- verified via `bal build` plus the full unit, client, and subscription test suites passing.

ballerina/client_subscription_connection.bal
ThisaruGuruge: Update: extracted the shared abstraction in 3ea27494. Both sides' keep-alive loop now live once in a new `keep_alive.bal`: a `KeepAlivePeer` isolated object type (`sendPing()`, `onKeepAliveTimeout()`, `isExternallyStopped()`) captures the one real difference between the two sides, with `ClientKeepAlivePeer` (wrapping `websocket:Client`, calling back into `SubscriptionConnection.handleKeepAliveTimeout`) and `ServerKeepAlivePeer` (wrapping `websocket:Caller`, closing via `closeConnection`) as the two adapters. `KeepAliveMonitor" and `keepAliveWait`/`runKeepAliveLoop` are now shared verbatim by both sides. Existing test coverage on both the client and server keep-alive paths passes unchanged.

native/src/main/java/io/ballerina/stdlib/graphql/runtime/client/SubscriptionConnectionState.java
ThisaruGuruge: Update: rewrote this as a proper state machine in ff601ad1. Replaced the `closed`/`connecting` `AtomicBoolean` pair with a single `AtomicReference<ConnectionLifecycleState>` (a 5-value enum: `DISCONNECTED`, `CONNECTING`, `CONNECTED`, `RECONNECTING`, `CLOSED`), with `CLOSED` enforced as a sticky/terminal state via a `transitionUnlessClosed` guard so a racing connect/reconnect attempt can never move the state back out of `CLOSED` once `close()` has run. Read every call site in `client_subscription_connection.bal` first rather than guessing at the transitions, then updated all of them (`isClosed`/`isConnecting`/`startConnecting`/`startReconnecting`/`markConnected`/`markDisconnected`). Native module rebuilt and the full test suite re-verified passing.

ballerina-tests/graphql-client-test-suite/tests/subscription_mock_services.bal
```diff
@@ -0,0 +1,499 @@
...
+    isolated resource function get .() returns websocket:Service|websocket:UpgradeError {
+        int connectionNumber;
+        lock {
+            mockExhaustionConnectionCount += 1;
+            connectionNumber = mockExhaustionConnectionCount;
+        }
+        if connectionNumber > 1 {
+            return error("Service unavailable", code = 503);
+        }
+        return new MockCloseReconnectWsService();
```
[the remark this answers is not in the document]
ThisaruGuruge: Confirmed still valid against the current code. Fixed in db7cc62e: `/mock_exhaustion` now uses a dedicated `MockExhaustionWsService` (same ping/init/subscribe-drop behavior) instead of `MockCloseReconnectWsService`, so it no longer touches the shared `mockCloseReconnectDropped` flag that `testCloseDuringReconnect` relies on. Verified both `testReconnectExhaustion` and `testCloseDuringReconnect` still pass (70/70 in the client test suite).

ballerina/client_subscription_connection.bal
```diff
@@ -0,0 +1,724 @@
...
+            string? id = getWsMessageId(message);
+            if id is () {
+                return;
+            }
+            if messageType == WS_NEXT {
+                // Messages received for an unknown ID are ignored.
+                MessageQueue? queue = self.getQueue(id);
+                if queue is MessageQueue {
+                    queue.enqueue(getWsMessagePayload(message));
+                }
```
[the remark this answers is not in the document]
ThisaruGuruge: Confirmed still valid -- this exact bug survived the dispatch-loop refactor unchanged (now at `client_subscription_connection.bal`'s `dispatchMessage`, `WS_NEXT` branch, around what is currently line 402-415). Fixed in 07d8b65b: a nil payload on a `next` message now enqueues `error SubscriptionError(INVALID_SUBSCRIPTION_MESSAGE, errors = ())` instead of the raw nil, so it can no longer be mistaken for `MessageQueue`'s completion sentinel. Verified via the full unit, client, and subscription test suites passing.

ballerina/Dependencies.toml
```diff
@@ -403,7 +403,7 @@ modules = [
 [[package]]
 org = "ballerina"
 name = "websocket"
-version = "2.15.0"
+version = "2.15.4"
```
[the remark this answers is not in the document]
ThisaruGuruge: Confirmed still valid -- `gradle.properties`' `stdlibWebsocketVersion` is 2.15.5 but `ballerina/Dependencies.toml` still recorded 2.15.4 for the `websocket` package. Fixed in b129347b by bumping just that recorded version to 2.15.5 (its dependency list is unchanged, so this matches what a full lockfile regeneration would produce for this entry). Deliberately did not pull in the broader dependency bumps (http, crypto, io, mime, task, time, url, data.jsondata) that a raw `bal build` also picks up from Ballerina Central right now -- those are unrelated to this skew and are better handled through the repo's normal stdlib dependency-bump process rather than as a side effect of a review-response commit.

docs/spec/spec.md
```diff
@@ -462,18 +548,85 @@ type Profile record {|
...
+###### Example: Subscription Configuration with Authentication and Reconnection
+
+```ballerina
+graphql:Client donationsClient = check new ("http://localhost:9090/donations",
+    subscription = {
+        connectionInitPayload: {authToken: token},
+        reconnect: {maxAttempts: 3, interval: 2},
+        websocketConfig: {
+            customHeaders: {"Authorization": string `Bearer ${token}`}
+        }
```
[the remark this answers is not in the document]
ThisaruGuruge: Confirmed still valid. Fixed in 716c848d: the authenticated subscription example now uses `https://localhost:9090/donations`, so the derived WebSocket connection is `wss://` rather than `ws://`.

ballerina/client_subscription_connection.bal
```diff
@@ -0,0 +1,820 @@
...
+        if self.keepAliveEnabled {
+            KeepAliveMonitor monitor = new;
+            keepAliveMonitor = monitor;
+            ClientKeepAlivePeer peer = new (wsClient, self);
+            _ = start runKeepAliveLoop(monitor, self.keepAliveSignalQueue, self.keepAlivePingInterval,
+                    self.keepAlivePongTimeout, peer);
+        }
+        while true {
+            json|websocket:Error message = wsClient->readMessage();
+            if message is websocket:ReadTimedOutError {
```
ThisaruGuruge, line 336: **Potential infinite loop / orphaned strand after `close()` (and after an abandoned `connect()` attempt).**

Both read loops in `dispatch()` treat `websocket:ReadTimedOutError` as an unconditional retry, with no check of `self.isClosed()` (or of whether this `wsClient` has since been replaced/abandoned):

```ballerina
json|websocket:Error message = wsClient->readMessage();
if message is websocket:ReadTimedOutError {
    // Idle read timeout, not a connection failure: retry the read.
    continue;
}
```

The sibling occurrence in the ack-wait loop is at line 289-292.

Per the commit that introduced this idle-read watchdog ("`Caller.close()`'s timeout-based termination isn't reliably honored" on Linux), a `wsClient->close()` call does not reliably surface as a `websocket:Error` on a concurrently-blocked `readMessage()`. Concretely:

1. A subscription connection is established; `dispatch()` is in the post-ack loop.
2. The user calls `client->close()` -> `SubscriptionConnection.close()` -> `closeLocked()` -> `wsClient->close(timeout = GRACEFUL_CLOSE_TIMEOUT)`. On the affected platform, the peer's close-frame echo doesn't arrive and the local read is not interrupted either.
3. The blocked `readMessage()` eventually raises `ReadTimedOutError` (bounded by `readTimeout`/`DEFAULT_IDLE_READ_TIMEOUT`), which is unconditionally retried here.
4. Nothing in this branch ever checks `self.isClosed()`, so on a socket that keeps failing to deliver a real error, this strand retries `readMessage()` forever - it never returns, never calls `handleConnectionFailure`, and is never told to stop (Ballerina has no strand-cancellation mechanism).

Since this is exactly the failure mode this fix targets (unreliable close termination), the fix doesn't fully close the gap - it just changes an indefinite block into a same-length-cycle infinite retry loop that leaks the strand instead of hanging on one read call. This looks like a plausible contributor to the CI-hang symptoms this branch has been chasing.

Suggested fix: check `self.isClosed()` (and/or that `self.loadWsClient() === wsClient`) on the `ReadTimedOutError` branch and `return` if so, in both loops.

ThisaruGuruge: Fixed in 0d959f69 — both retry branches (this one and the ack-wait loop at line 289-292) now bail out on `self.isClosed()` (and, in this main dispatch loop, also `self.loadWsClient() !== wsClient` to catch a superseded reconnect) instead of retrying forever on a dead socket. Verified locally: `graphql-client-test-suite` now passes in full, including `testPingPongKeepAlive` and `testSubscriptionWithGenericResponseBinding`, which were failing with a keep-alive `CompleteEarlyError` on the last CI run before this fix.

gayaldassanayake (approved):

TharmiganK (approved):

